PR #3, opened 2017-04-07
Support bundled libffi

unak:
Just like current ruby trunk


hsbt:
Thank you! I will setup appveyor build with this pull request later.

Issue #7, opened 2017-11-12
How to import Kernel32 function 'GetVolumePathName', with Windows types?


rg-3:
The code:
```ruby
module WFFI
  require 'fiddle'
  require 'fiddle/import'
  require 'fiddle/types'
  extend Fiddle::Importer
  dlload 'kernel32'
  include Fiddle::Win32Types
  extern "BOOL WINAPI GetVolumePathName(LPCTSTR lpszFileName, LPTSTR lpszVolumePathName, DWORD cchBu\
fferLength);"
end
```

Error:
```
C:/tools/ruby24/lib/ruby/2.4.0/fiddle/cparser.rb:177:in `parse_ctype': unknown type: LPCTSTR (Fiddle::DLError)
        from C:/tools/ruby24/lib/ruby/2.4.0/fiddle/cparser.rb:90:in `block in parse_signature'
        from C:/tools/ruby24/lib/ruby/2.4.0/fiddle/cparser.rb:90:in `collect'
        from C:/tools/ruby24/lib/ruby/2.4.0/fiddle/cparser.rb:90:in `parse_signature'
        from C:/tools/ruby24/lib/ruby/2.4.0/fiddle/import.rb:166:in `extern'
        from fiddle.rb:10:in `<module:WFFI>'
        from fiddle.rb:3:in `<main>'
```

I will try with WIN32API now since I cannot discover the solution so far.

__Notes:__
There's a number of errors that can happen before I got this far:
 
* If you don't 'extend FFI::Importer' first, `include Fiddle::Win32Types` raises an error (`NoMethodError typealias`). I have a suggestion: Fiddle::Win32Importer. 
It is Win32Types+Importer module. I can PR if you like the idea.

* If you don't call `dlload` before `include Fiddle::Win32Types`, there is another NoMethodError. I think point mentioned above would solve also. Remove the chances of error while using this library.

Thanks for your work!

sancarn:
[Stack overflow](https://stackoverflow.com/questions/321413/lpcstr-lpctstr-and-lptstr) says that `LPCTSTR` is a `TCHAR` string.

PR #8, opened 2017-11-12
address some concerns in #7

rg-3:
Just an idea. There could be User32 importer and maybe some other common DLLs.
Does it work? Thanks.

### DIFF
--- a/lib/fiddle/win32.rb
+++ b/lib/fiddle/win32.rb
@@ -1,0 +1,2 @@
+module Fiddle::Win32
+end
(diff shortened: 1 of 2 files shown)


r-obert:
Some alternative names:
```ruby
Fiddle::Importer::Kernel32
Fiddle::Importer::Win32::Kernel32
```

r-obert:
Alternative:
```ruby
module Fiddle::Importer::Win32
  def self.kernel32
    Module.new do
      extend Fiddle::Importer
      dlload "kernel32"
      include Fiddle::Win32Types
    end
end

module WinFFI
  extend Fiddle::Importer::Win32.kernel32
end
```

Just some ideas.

PR #13, opened 2018-10-29
Fix a failing test

mrkn:
This commit fixes the following failure:

```
  1) Failure:
Fiddle::TestImport#test_no_message_with_debug [/Users/mrkn/src/github.com/ruby/fiddle/test/fiddle/test_import.rb:152]:

1. [2/2] Assertion for "stderr"
   | <[]> expected but was
   | <["Exception `NameError' at /Users/mrkn/.rbenv/versions/2.5.1/lib/ruby/2.5.0/fiddle/import.rb:157 - uninitialized constant Fiddle::Function::STDCALL"]>.
```

### DIFF
--- a/test/fiddle/test_import.rb
+++ b/test/fiddle/test_import.rb
@@ -149,16 +149,13 @@ module Fiddle
     end
 
     def test_no_message_with_debug
-      # disable bundler
-      orig_RUBYOPT = ENV['RUBYOPT']
-      ENV['RUBYOPT'] = orig_RUBYOPT.gsub(%r|-rbundler/setup|, '') if orig_RUBYOPT
+      # disable all Ruby environment variables
+      orig_RUBYOPT, ENV['RUBYOPT'] = ENV['RUBYOPT'], nil
+      orig_RUBYLIB, ENV['RUBYLIB'] = ENV['RUBYLIB'], nil
 
       # load development fiddle instead of bundled one
-      orig_RUBYLIB = ENV['RUBYLIB']
       libdir = File.expand_path('../../../lib', __FILE__)
-      if File.file?(File.join(libdir, "fiddle/import.rb"))
-        ENV['RUBYLIB'] = [libdir, orig_RUBYLIB].compact.join(":")
-      end
+      ENV['RUBYLIB'] = libdir if File.file?(File.join(libdir, "fiddle/import.rb"))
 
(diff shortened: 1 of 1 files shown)


mrkn:
OMG

nobu:
Seems the bundled fiddle is called.

nobu:
I suspect that this failure is caused by a bug in bundler; bundler/setup.rb enables bundler even if rubygems is disabled.

mrkn:
@nobu I just remove `-rbundler/setup` from `RUBYOPT` instead of using `Bundler.with_clean_env`.

mrkn:
Oh...

kou:
How about disabling all `RUBYOPT` and `RUBYLIB` in this test? It's simpler and more stable.

This works even if a developer has the following ugly `RUBYOPT`:

```console
$ cat /tmp/warn.rb
$, = "XXX"
$ RUBYOPT=-r/tmp/warn.rb rake
```

```diff
diff --git a/test/fiddle/test_import.rb b/test/fiddle/test_import.rb
index 99294ea..db58a82 100644
--- a/test/fiddle/test_import.rb
+++ b/test/fiddle/test_import.rb
@@ -149,7 +149,21 @@ module Fiddle
     end
 
     def test_no_message_with_debug
-      assert_in_out_err(%w[--debug --disable=gems -rfiddle/import], 'p Fiddle::Importer', ['Fiddle::Importer'])
+      # disable all Ruby environment variables
+      orig_RUBYOPT, ENV['RUBYOPT'] = ENV['RUBYOPT'], nil
+      orig_RUBYLIB, ENV['RUBYLIB'] = ENV['RUBYLIB'], nil
+      libdir = File.expand_path('../../../lib', __FILE__)
+      assert_in_out_err([
+                          "--debug",
+                          "--disable=gems",
+                          "-I#{libdir}",
+                          "-rfiddle/import",
+                        ],
+                        'p Fiddle::Importer',
+                        ['Fiddle::Importer'])
+    ensure
+      ENV['RUBYLIB'] = orig_RUBYLIB
+      ENV['RUBYOPT'] = orig_RUBYOPT
     end
   end
 end if defined?(Fiddle)
```

mrkn:
@kou thanks for your advice.
I pushed the new commit that follows your patch.

nobu:
The first element of `args` to `assert_in_out_err` can be a `Hash` as environment variables.

kou:
!!!
I've changed to use it.

Issue #17, opened 2019-05-27
Implement buffer protocol for Ruby


kou:
See also PEP-3118 -- Revising the buffer protocol
https://www.python.org/dev/peps/pep-3118/

ioquatix:
You want to implement support for python buffers in Ruby?

kou:
No.
It's for Ruby. It doesn't care about compatibility with Python's buffer protocol (at least for now).
Apache Arrow, XND or something will be suitable for data exchange over languages.

ioquatix:
How does it fit with existing I/O operations? Can we use this new proposed buffer instead of strings for I/O?

kou:
I don't think about I/O related operations for the buffer protocol. I just think about data exchange.
It may be interesting. Do you have any (potential) use-case with I/O operations? Async I/O related?

ioquatix:
It would be great to support `readv` and `writev` by supplying array rather than string. Maybe buffer operation can do the same. It would be nice to handle explicit offsets into buffers for both `read` and `write` operations, e.g. <https://bugs.ruby-lang.org/issues/15560>.

There are issues with handling binary string concatenation, e.g. <https://bugs.ruby-lang.org/issues/14975>.

We end up treating a string like a binary buffer, but it's often the case this duality causes weird interface issues, e.g. <https://bugs.ruby-lang.org/issues/13626>.

If we can introduce some kind of buffer concept, with mapping to string, and support for multiple chunks, to make I/O operation more efficient, it would be fantastic. But it needs to be built with I/O `readv` and `writev` in mind.

kou:
Thanks.
I'll read them later.

dsisnero:
I added this to developers meeting- Hopefully it gets accepted https://bugs.ruby-lang.org/issues/16933

kou:
It's discussing in ruby/ruby#3261 .

PR #18, opened 2019-05-31
WIP: Add buffer protocol

mrkn:
This implements Ruby-version of PEP-3118 (#17 ).


mrkn:
Buffer protocol for Ruby will be implemented in CRuby's core.
See https://github.com/ruby/ruby/pull/3261

PR #19, opened 2019-06-13
Use RbConfig::SIZEOF

nobu:
The following code is wrong some platforms, e.g., mswin/mingw.

```ruby
  case [0].pack('L!').size
  when 4
    # 32-bit ruby
  when 8
    # 64-bit ruby
  end
```


kou:
These codes are used on only Linux and Solaris. But this is a good manner.

PR #23, opened 2020-01-05
Add Fiddle.ruby_debug_breakpoint

mrkn:
libruby.so has `ruby_debug_breakpoint` function that can be used for inserting a breakpoint when debugging on C-layer.
I sometimes need to call this function from Ruby-layer to enter the gdb interactive session.
So I'd like to add `Fiddle.ruby_debug_breakpoint`.

### DIFF
--- a/lib/fiddle/breakpoint.rb
+++ b/lib/fiddle/breakpoint.rb
@@ -1,0 +1,10 @@
+require "fiddle"
+
+module Fiddle
+  @ruby_debug_breakpoint = Function.new(Handle.sym("ruby_debug_breakpoint"),
+                                        [], Fiddle::TYPE_VOID)
+
+  def self.ruby_debug_breakpoint
+    @ruby_debug_breakpoint.()
+  end
+end


lib/fiddle/breakpoint.rb
```diff
@@ -0,0 +1,10 @@
+require "fiddle"
+
+module Fiddle
+  @ruby_debug_breakpoint = Function.new(Handle.sym("ruby_debug_breakpoint"),
+                                        [], Fiddle::TYPE_VOID)
+
+  def self.ruby_debug_breakpoint
```
mrkn: I don't think the name of this method is good.
Any candidate is welcome.

ko1:
* `ruby_debug_breakpoint` seems not exposed, doesn't it?
* `ruby_debug_breakpoint` is only internal API so there is no discussion to introduce it. I'm afraid it is good name and functionality.

kou:
You can use `Process.kill(:TRAP, Process.pid)` for the purpose.
Do you still need this API?

PR #14, opened 2018-10-30
Add support for nested structs

sinisterchipmunk:
This pull request does a few things. The main goal of the PR is to add support for nested structs and unions containing structs, as discussed in #12. I need this for a non-trivial project and have been using the nested struct code in numerous places within that project for several weeks now, so I am confident it is working properly. The syntax is a little strange, but works quite well, and is easy to learn. One could express the example given in #12 like so:

```ruby
StudentCollegeDetail = struct [
  'int college_id',
  'char college_name[50]'
]

StudentDetail = struct [
  'int id',
  'char name[20]',
  { clg_data: StudentCollegeDetail }
]
```

It is also possible to specify the nested struct in-line, by replacing `StudentCollegeDetail` with `struct([...])`, like so: 

```ruby
StudentDetail = struct [
  'int id',
  'char name[20]',
  {
    clg_data: struct([
                      'int college_id',
                      'char college_name[50]'
                    ])
  }
]
```

The hash does not need to appear last. In fact, its position in the list of members dictates the offset into the parent struct at which the nested struct appears, in order to play nicely with C structs. There can be multiple hashes, or a single hash with multiple keys/values, or any combination of these. The following examples are also valid syntax, but will result in the struct members being laid out at different offsets within the outer struct's memory:

```ruby
# order of members in memory: position, id, dimensions
Rect = struct [ { position: struct(['float x', 'float y']) },
                'int id',
                { dimensions: struct(['float w', 'float h']) }
              ]

# order of members in memory: id, position, dimensions (assuming ordered hash)
Rect = struct [ 'int id',
                {
                  position: struct(['float x', 'float y']),
                  dimensions: struct(['float w', 'float h'])
                }
              ]
```

Second: when calling `Fiddle.malloc`, the contents of the resulting memory were not initialized to 0 as they were when calling `Fiddle::Pointer.malloc`. This inconsistency caused me a few headaches so I decided to fix it.

Third: `CStructEntity` defines array set and get methods (`[]` / `[]=`), which is fine except that they override `Fiddle::Pointer#[]` and `Fiddle::Pointer#[]=` and do not handle the 3rd argument that the base class allows for. I found that if I wanted to access the underlying memory of a struct directly, instead of working with the struct members, I had no good way to do that. So I modified these methods to check for an additional argument, and if present, delegate that to `super` so that the underlying memory could be accessed / assigned directly.

[...]

PR #26, opened 2020-01-22
Fix assignment to array within struct

sinisterchipmunk:
This pull request was split from #14 as requested by @nobu . It depends on #25 because that pull request makes it possible to access the underlying memory of a struct, which is required in order for this pull request to work properly.

When a member of a struct is an array, `CStructEntity` would return a temporary array containing those values. Assignment to that array was a no-op, which led to confusing behavior because an assignment to the array would be "ignored". I fixed this so that assignment to the returned array correctly updates the member within the struct.

For example:

```ruby
a = struct(['int i[5]']).malloc
a.i[2] = 1 # previously, this would do nothing
a.i[2] #=> 1
```

### DIFF
--- a/lib/fiddle/struct.rb
+++ b/lib/fiddle/struct.rb
@@ -39,7 +39,7 @@ module Fiddle
 
     def []=(index, value)
       if index < 0 || index >= size
-        raise RangeError, 'index %d outside of array bounds 0...%d' % [index, size]
+        raise IndexError, 'index %d outside of array bounds 0...%d' % [index, size]
       end
 
       to_ptr[index * @align, @size] = [value].pack(@pack_format)
--- a/lib/fiddle/struct.rb
+++ b/lib/fiddle/struct.rb
@@ -149,7 +149,9 @@ module Fiddle
     end
 
     # Fetch struct member +name+
-    def [](name)
+    def [](*args)
+      return super(*args) if args.size > 1
+      name = args[0]
       idx = @members.index(name)
       if( idx.nil? )
         raise(ArgumentError, "no such member: #{name}")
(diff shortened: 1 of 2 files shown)


lib/fiddle/struct.rb
```diff
@@ -20,6 +20,33 @@ def CUnion.entity_class
...
+      super(initial_values.collect { |v| unsigned_value(v, type) })
+    end
+
+    def to_ptr
+      @ptr
+    end
+
+    def []=(index, value)
+      if index < 0 || index >= size
+        raise RangeError, 'index %d outside of array bounds 0...%d' % [index, size]
```
nobu: Maybe `IndexError`?

lib/fiddle/struct.rb
```diff
@@ -20,6 +20,33 @@ def CUnion.entity_class
...
+      end
+
+      to_ptr[index * @align, @size] = [value].pack(@pack_format)
```
nobu: Why `@align`, shouldn't be `@size`?

sinisterchipmunk: I agree this was wrong. I fixed the error, but I was unable to make it fail in a test case.

The test did not fail because, on my system, `@align == @size` for all primitive types. Therefore the result of using `@align` yields a correct value on my system, but could fail on other architectures. I modified the test case to attempt to write at several array indices, and then check that the underlying memory can be correctly unpacked. I think the modified test case covers the issue better than the original, even though I did not experience a failure, so I committed it.

lib/fiddle/struct.rb
sinisterchipmunk: Good idea, I changed this.

PR #33, opened 2020-05-18
Improve documentation on how to correctly free memory and free memory in tests

chrisseaton:
The documentation for pointers and structs doesn't include as much helpful information as it could about who is responsible for freeing memory and how to do it safely. I've added some specific examples for guidance.

I've then also applied this in the test suite, which I think was not freeing memory in many tests.

Or am I misunderstanding how freeing pointers and structs works in Fiddle?

If I'm understanding it all correctly, I'd then like to follow this up with some additions to the API to make it even easier to do the right thing, such as

```ruby
Fiddle::Pointer.malloc(size) do |pointer|
   ...
end
# memory freed when coming out of scope
```

```ruby
pointer = Fiddle::Pointer.malloc(size)
pointer.free_now
# calls Fiddle.free
```

```ruby
pointer = Fiddle::Pointer.malloc(size, free_routine)
pointer.free_now
# calls Fiddle.free and also de-registers the `freefunc`, for manual freeing but also GC as a backup
```

### DIFF
--- a/lib/fiddle/struct.rb
+++ b/lib/fiddle/struct.rb
@@ -73,7 +73,12 @@ module Fiddle
     #
     #   MyStruct = Fiddle::CStructBuilder.create(Fiddle::CUnion, types, members)
     #
-    #   obj = MyStruct.allocate
+    #   obj = MyStruct.malloc
+    #   begin
+    #     ...
+    #   ensure
+    #     Fiddle.free obj.to_ptr
+    #   end
     #
     def create(klass, types, members)
       new_class = Class.new(klass){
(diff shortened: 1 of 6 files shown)


lib/fiddle/struct.rb
```diff
@@ -73,7 +73,12 @@ module CStructBuilder
     #
     #   MyStruct = Fiddle::CStructBuilder.create(Fiddle::CUnion, types, members)
     #
-    #   obj = MyStruct.allocate
+    #   obj = MyStruct.malloc
```
chrisseaton, line 76: I don't think `allocate` was ever correct?

tenderlove:
Adding the block API makes sense to me

PR #35, opened 2020-05-26
Add GH Actions

MSP-Greg:
Today the ruby head mswin build broke from fiddle.  ruby/ruby appears to download libffi 3.2.1 for the mswin build, but 3.3 is vendored here.

I started here and created a GitHub Actions workflow, covering Ruby 2.3 thru master/head for Ubuntu, macOS, and Windows.  All are passing except Windows mswin.   I did add a compile step just to see what was happening.

This may not run, as often the first PR with a GHA workflow doesn't.  A link to a run in my fork is:
https://github.com/MSP-Greg/fiddle/actions/runs/116195157

### DIFF
--- a/.github/workflows/fiddle.yml
+++ b/.github/workflows/fiddle.yml
@@ -1,0 +1,42 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  CI:
+    name: >-
+      ${{ matrix.os }} ${{ matrix.ruby }}
+    runs-on: ${{ matrix.os }}-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu, macos, windows ]
+        ruby: [ 2.3, 2.4, 2.5, 2.6, 2.7, head ]
+        include:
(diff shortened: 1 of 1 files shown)


kou:
It seems that "windows mswin" is failed: https://github.com/MSP-Greg/fiddle/runs/711008486?check_suite_focus=true

What should we do for this?

kou:
Ah, you want to show "windows mswin" is failed.

kou:
Should we merge this and fix the problem?

MSP-Greg:
> Ah, you want to show "windows mswin" is failed.

Yes.  Sorry, I should have opened an issue for that, rather than noting it in this PR.  Also, the issue with ext/fiddle/extlibs using 3.2.1 both here and in ruby/ruby.

> Should we merge this and fix the problem?

Re merging, whatever you prefer.  The CI runs quickly, so I used a 'complete' matrix of jobs.  As to fixing the problem, my c knowledge is limited, so I wouldn't be much help.  Thanks.

kou:
OK.
I'll merge this.

PR #36, opened 2020-05-31
Add a Fiddle::Pointer#call_free method

chrisseaton:
`free!` runs the user's free function. It doesn't do anything if the pointer has already been freed. The GC won't call the user's free function if the user already manually freed. Another method `freed?` tells the user if the pointer has already been freed.

All together, these changes allow you to rely on the GC to definitely free your memory, meaning no leaks, but then also allows you to manually free as early as possible if you can, meaning deterministic deallocation, improving cache locality and and reducing peaks in memory use.

### DIFF
--- a/ext/fiddle/pointer.c
+++ b/ext/fiddle/pointer.c
@@ -232,7 +232,7 @@ rb_fiddle_ptr_initialize(int argc, VALUE argv[], VALUE self)
  * if not already manually called via *call_free*.
  *
  * +freefunc+ must be an address pointing to a function or an instance of
- * +Fiddle::Function+. Using +freefunc+ without using +free!+ may lead to
+ * +Fiddle::Function+. Using +freefunc+ without using +call_free+ may lead to
  * unlimited memory being allocated before any is freed as the native memory
  * the pointer references does not contribute to triggering the Ruby garbage
  * collector. Consider manually freeing the memory as illustrated above. You
--- a/ext/fiddle/pointer.c
+++ b/ext/fiddle/pointer.c
@@ -63,9 +63,9 @@ fiddle_ptr_free_ptr(void *ptr)
 {
     struct ptr_data *data = ptr;
     if (data->ptr && data->free && !data->freed) {
+	data->freed = true;
 	(*(data->free))(data->ptr);
     }
-    data->freed = true;
 }
 
 static void
--- a/ext/fiddle/pointer.c
+++ b/ext/fiddle/pointer.c
@@ -387,13 +387,13 @@ rb_fiddle_ptr_free_get(VALUE self)
 }
 
 /*
- * call-seq: free! => nil
+ * call-seq: call_free => nil
  *
  * Call the free function for this pointer. Calling more than once will do
  * nothing. Does nothing if there is no free function attached.
  */
 static VALUE
-rb_fiddle_ptr_free_bang(VALUE self)
+rb_fiddle_ptr_call_free(VALUE self)
 {
     struct ptr_data *pdata;
     TypedData_Get_Struct(self, struct ptr_data, &fiddle_ptr_data_type, pdata);
@@ -755,7 +755,7 @@ Init_fiddle_pointer(void)
     rb_define_method(rb_cPointer, "initialize", rb_fiddle_ptr_initialize, -1);
     rb_define_method(rb_cPointer, "free=", rb_fiddle_ptr_free_set, 1);
     rb_define_method(rb_cPointer, "free",  rb_fiddle_ptr_free_get, 0);
-    rb_define_method(rb_cPointer, "free!",  rb_fiddle_ptr_free_bang, 0);
+    rb_define_method(rb_cPointer, "call_free",  rb_fiddle_ptr_call_free, 0);
     rb_define_method(rb_cPointer, "freed?",  rb_fiddle_ptr_freed_p, 0);
     rb_define_method(rb_cPointer, "to_i",  rb_fiddle_ptr_to_i, 0);
     rb_define_method(rb_cPointer, "to_int",  rb_fiddle_ptr_to_i, 0);
--- a/ext/fiddle/pointer.c
+++ b/ext/fiddle/pointer.c
@@ -229,7 +229,7 @@ rb_fiddle_ptr_initialize(int argc, VALUE argv[], VALUE self)
  *
  * Allocate +size+ bytes of memory and associate it with an optional
  * +freefunc+ that will be called when the pointer is garbage collected,
- * if not already manually called via *free!*.
+ * if not already manually called via *call_free*.
  *
  * +freefunc+ must be an address pointing to a function or an instance of
  * +Fiddle::Function+. Using +freefunc+ without using +free!+ may lead to
--- a/ext/fiddle/pointer.c
+++ b/ext/fiddle/pointer.c
@@ -209,7 +209,7 @@ rb_fiddle_ptr_initialize(int argc, VALUE argv[], VALUE self)
  *    # Manually freeing but relying on the garbage collector otherwise
  *    pointer = Fiddle::Pointer.malloc(size, Fiddle::RUBY_FREE)
  *    ...
- *    pointer.free!
+ *    pointer.call_free
  *
  *    # Relying on the garbage collector - may lead to unlimited memory allocated before freeing any, but safe
  *    pointer = Fiddle::Pointer.malloc(size, Fiddle::RUBY_FREE)
--- a/ext/fiddle/pointer.c
+++ b/ext/fiddle/pointer.c
@@ -24,6 +24,7 @@ struct ptr_data {
     void *ptr;
     long size;
     freefunc_t free;
+    int freed;
     VALUE wrap[2];
 };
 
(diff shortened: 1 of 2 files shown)


ext/fiddle/pointer.c
```diff
@@ -24,6 +24,7 @@ struct ptr_data {
     void *ptr;
     long size;
     freefunc_t free;
+    int freed;
```
chrisseaton, line 27: Is `int` and `1` and `0` the right way to do booleans in C extensions? I found some other things but wasn't sure if they were internal for MRI?

chrisseaton:
Some of these Windows CI failures don't make sense to me - some say they can't find `libffi` and some look like they're testing the *system* Fiddle, not the Fiddle in this repository.

test/fiddle/test_pointer.rb
```diff
@@ -170,6 +186,31 @@ def test_free=
       assert_equal free.ptr, ptr.free.ptr
     end
 
+    def test_free!
+      ptr = Pointer.malloc(4)
+      refute ptr.freed?
+      ptr.free!
+      assert ptr.freed?
+      ptr.free!                 # you can safely run it again
+      assert ptr.freed?
```
chrisseaton, line 195: What do we think `free!` and `freed?` should do when there's no free function attached?

test/fiddle/test_pointer.rb
```diff
@@ -20,16 +20,28 @@ def test_malloc_free_func_int
       assert_equal free.to_i, Fiddle::RUBY_FREE.to_i
 
       ptr  = Pointer.malloc(10, free.to_i)
-      assert_equal 10, ptr.size
-      assert_equal free.to_i, ptr.free.to_i
+      refute ptr.freed?
+      begin
+        assert_equal 10, ptr.size
+        assert_equal free.to_i, ptr.free.to_i
+      ensure
```
kou, line 27: Why do we need this?
The `ptr` is freed by GC when on of the above two assertions is failed,  isn't it?

test/fiddle/test_pointer.rb
```diff
@@ -20,16 +20,28 @@ def test_malloc_free_func_int
       assert_equal free.to_i, Fiddle::RUBY_FREE.to_i
 
       ptr  = Pointer.malloc(10, free.to_i)
-      assert_equal 10, ptr.size
-      assert_equal free.to_i, ptr.free.to_i
+      refute ptr.freed?
```
kou, line 23: Do we need test `freed?` here?
It seems that you add separated tests for it.

test/fiddle/test_pointer.rb
kou: How about doing nothing?

```ruby
refute ptr.freed?
ptr.free!
refute ptr.freed?
ptr.free!
refute ptr.freed?
```

test/fiddle/test_pointer.rb
```diff
@@ -170,6 +186,31 @@ def test_free=
...
+      assert ptr.freed?
+
+      ptr = Pointer.malloc(4, Fiddle::RUBY_FREE)
+      refute ptr.freed?
+      ptr.free!
+      assert ptr.freed?
+      ptr.free!                 # you can safely run it again
+      assert ptr.freed?
+      GC.start                  # you can safely run the GC routine
+      assert ptr.freed?
```
kou, line 204: Could you split this case to another test?

ext/fiddle/pointer.c
```diff
@@ -370,6 +385,34 @@ rb_fiddle_ptr_free_get(VALUE self)
     return rb_fiddle_new_function(address, arg_types, ret_type);
 }
 
+/*
+ * call-seq: free! => nil
```
kou, line 389: How about renaming to `call_free` or something?
`free` (getter) and `free!` will confuse us.

kou:
We can also ignore all Windows jobs...

ext/fiddle/pointer.c
chrisseaton: This failed in all CI configurations when I tried it.

test/fiddle/test_pointer.rb
chrisseaton: Yes that's what we do at the moment. As long as everyone else also thinks that's fine I'll do that.

test/fiddle/test_pointer.rb
chrisseaton: In my mind, it's good practice to free a native, external resource as soon as you can. That's why we use blocks for example when opening a file, rather than letting the GC close it. In the future I hope to create a PR for a similar block scope for pointers as well.

ext/fiddle/pointer.c
kou: We need `#include <stdbool.h>`.

ext/fiddle/pointer.c
chrisseaton: Yes... I was going for terseness, but I'm open to debating this.

chrisseaton: It's also a destructive operation, which fits the `!` pattern.

kou: `free!` is destructive but `free!` is not the destructive version of `free`.
The relation of them is different from the relation of `String#gsub` and `String#gsub!`.

MSP-Greg:
@chrisseaton @kou

I apologize for not looking at the Windows logs in Actions.

All Windows Ruby versions, 2.4 and later, have a vendored file, `rubygems/defaults/operating_system.rb`, that then requires files that eventually require fiddle.  It's used for a win32 call that adds folders for dll resolution without adding the folder to PATH.

See https://github.com/ruby/fiddle/pull/37, which disables the loading of `operating_system.rb` via `--disable=gems`...

ext/fiddle/pointer.c
chrisseaton: Done.

test/fiddle/test_pointer.rb
chrisseaton: Done.

test/fiddle/test_pointer.rb
chrisseaton: Tidied up.

test/fiddle/test_pointer.rb
chrisseaton: Done.

ext/fiddle/pointer.c
chrisseaton: I feel like `free!` is best due to being short and looking destructive. I'm open to changing it if you're very certain or other people agree.

kou: If we have block style free feature (`Fiddle::Pointer.malloc(size, Fiddle::RUBY_FREE) {|pointer| ...}`?), we'll not call this method manually.
So I don't think we should use shorter name for this method with potential confusion. (Some users may use `pointer.free` wrongly to free a pointer but the pointer isn't freed. It causes a memory leak.)

test/fiddle/test_pointer.rb
chrisseaton: Remove the manual free, leave it relying on GC for now, and then switch to the block malloc that's coming in the next PR? Ok.

ext/fiddle/pointer.c
chrisseaton: Ok I'll use `call_free`.

chrisseaton: Done.

test/fiddle/test_pointer.rb
chrisseaton: Done.

kou: Yes.

ext/fiddle/pointer.c
```diff
@@ -197,11 +206,16 @@ rb_fiddle_ptr_initialize(int argc, VALUE argv[], VALUE self)
  *
  * == Examples
  *
+ *    # Manually freeing but relying on the garbage collector otherwise
+ *    pointer = Fiddle::Pointer.malloc(size, Fiddle::RUBY_FREE)
+ *    ...
+ *    pointer.free!
```
kou, line 212:
```suggestion
 *    pointer.call_free
```

ext/fiddle/pointer.c
```diff
@@ -214,13 +228,15 @@ rb_fiddle_ptr_initialize(int argc, VALUE argv[], VALUE self)
  *    ...
  *
  * Allocate +size+ bytes of memory and associate it with an optional
- * +freefunc+ that will be called when the pointer is garbage collected.
+ * +freefunc+ that will be called when the pointer is garbage collected,
+ * if not already manually called via *free!*.
```
kou, line 232:
```suggestion
 * if not already manually called via *call_free*.
```

ext/fiddle/pointer.c
```diff
@@ -214,13 +228,15 @@ rb_fiddle_ptr_initialize(int argc, VALUE argv[], VALUE self)
...
+ * +freefunc+ that will be called when the pointer is garbage collected,
+ * if not already manually called via *call_free*.
+ *
  * +freefunc+ must be an address pointing to a function or an instance of
- * +Fiddle::Function+. Using +freefunc+ may lead to unlimited memory being
- * allocated before any is freed as the native memory the pointer references
- * does not contribute to triggering the Ruby garbage collector. Consider
- * manually freeing the memory as illustrated above. You cannot combine
- * the techniques as this may lead to a double-free.
+ * +Fiddle::Function+. Using +freefunc+ without using +free!+ may lead to
```
kou, line 235:
```suggestion
 * +Fiddle::Function+. Using +freefunc+ without using +call_free+ may lead to
```

kou:
Thanks!

PR #37, opened 2020-06-01
Fix Actions - test step - --disable=gems

MSP-Greg:
Since we're testing a default gem, disabling RubyGems is probably a good idea, especially since the test files are vendored.

Also, MinGW build's devkit requires fiddle, so that causes issues.

### DIFF
--- a/.github/workflows/fiddle.yml
+++ b/.github/workflows/fiddle.yml
@@ -37,6 +37,8 @@ jobs:
         run:  rake install
 
       - name: rake test
+        env:
+          RUBYOPT: --disable=gems
         run:  |
           ruby -v
           rake test


.github/workflows/fiddle.yml
```diff
@@ -37,6 +37,8 @@ jobs:
         run:  rake install
 
       - name: rake test
+        env:
+          rubyopt: --disable=gems
```
kou, line 41: Why is this `rubyopt` not `RUBYOPT`?

MSP-Greg: I can change it.  Windows ENV 'keys' are case-insensitive.  So, it only affects Windows builds.

kou: OK. Could you use `RUBYOPT` to avoid confusion?

MSP-Greg: Yes and done.

PR #38, opened 2020-06-03
Structured memory deallocation using blocks

chrisseaton:
```ruby
Fiddle::Pointer.new(size, free) do |pointer|
  ...
end
```

* Easy to use best-practice
* Helps people write correct code for the most common cases
* Protect against leaks (free if there's an exception, or even if someone uses a trace point to get hold of the object)
* Protect against deallocation latency (free as soon as you're done)
* Protect against non-locality (free while it's still in cache from your using it)

### DIFF
--- a/ext/fiddle/pointer.c
+++ b/ext/fiddle/pointer.c
@@ -272,8 +272,8 @@ rb_fiddle_ptr_s_malloc(int argc, VALUE argv[], VALUE klass)
     if (wrap) RPTR_DATA(obj)->wrap[1] = wrap;
 
     if (rb_block_given_p()) {
-        if (f == NULL) {
-            rb_raise(rb_eFiddleError, "a free function must be supplied to Fiddle::Pointer.malloc when it is called with a block");
+        if (!f) {
+            rb_raise(rb_eArgError, "a free function must be supplied to Fiddle::Pointer.malloc when it is called with a block");
         }
         return rb_ensure(rb_yield, obj, rb_fiddle_ptr_call_free, obj);
     } else {
--- a/ext/fiddle/pointer.c
+++ b/ext/fiddle/pointer.c
@@ -200,11 +200,20 @@ rb_fiddle_ptr_initialize(int argc, VALUE argv[], VALUE self)
     return Qnil;
 }
 
+static VALUE
+rb_fiddle_ptr_call_free(VALUE self);
+
 /*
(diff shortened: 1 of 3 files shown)


ext/fiddle/pointer.c
```diff
@@ -261,7 +271,14 @@ rb_fiddle_ptr_s_malloc(int argc, VALUE argv[], VALUE klass)
     obj = rb_fiddle_ptr_malloc(s,f);
     if (wrap) RPTR_DATA(obj)->wrap[1] = wrap;
 
-    return obj;
+    if (rb_block_given_p()) {
+        if (f == NULL) {
+            rb_raise(rb_eFiddleError, "a free function must be supplied to Fiddle::Pointer.malloc when it is called with a block");
```
kou, line 276: Could you use `rb_eArgError`?

ext/fiddle/pointer.c
```diff
@@ -261,7 +271,14 @@ rb_fiddle_ptr_s_malloc(int argc, VALUE argv[], VALUE klass)
     obj = rb_fiddle_ptr_malloc(s,f);
     if (wrap) RPTR_DATA(obj)->wrap[1] = wrap;
 
-    return obj;
+    if (rb_block_given_p()) {
+        if (f == NULL) {
```
kou, line 275:
```suggestion
        if (!f) {
```

Because we don't use `== NULL` in other places.

test/fiddle/test_pointer.rb
```diff
@@ -32,6 +32,24 @@ def test_malloc_free_func
...
+        escaped_ptr = ptr
+        :returned
+      end
+      assert_equal :returned, returned
+      assert escaped_ptr.freed?
+    end
+
+    def test_malloc_block_no_free
+      assert_raise Fiddle::DLError do 
+        Pointer.malloc(10) { |ptr| raise }
```
kou, line 49: Could you remove needless `raise`? `raise` in `assert_raise` may confuse us.

```suggestion
        Pointer.malloc(10) { |ptr| }
```

ext/fiddle/pointer.c
chrisseaton: Done.

chrisseaton:
Sorry I didn't use 'commit suggestion' I hadn't seen that before in GitHub and didn't realise it was an option.

PR #41, opened 2020-06-17
Add new pointer memory management interfaces to structs and unions

chrisseaton:
Allows you to go from

```ruby
struct = CStructEntity.malloc([TYPE_INT, TYPE_LONG])
begin
  ...
ensure
  Fiddle.free struct
end
```

to

```ruby
CStructEntity.malloc([TYPE_INT, TYPE_LONG], Fiddle::RUBY_FREE) do |struct|
  ..
end
```

and from

```ruby
timeval = LIBC::Timeval.malloc()
begin
  timezone = LIBC::Timezone.malloc()
  begin
    LIBC.gettimeofday(timeval, timezone)
  ensure
    Fiddle.free timezone.to_ptr
  end
ensure
  Fiddle.free timeval.to_ptr
end
```

to

```ruby
LIBC::Timeval.malloc(Fiddle::RUBY_FREE) do |timeval|
  LIBC::Timezone.malloc(Fiddle::RUBY_FREE) do |timezone|
    LIBC.gettimeofday(timeval, timezone)
  end
end
```

Both are safer, have better temporal and spatial locality for the memory manager, and reduce the working set size.

### DIFF
--- a/lib/fiddle/struct.rb
+++ b/lib/fiddle/struct.rb
@@ -62,7 +62,7 @@ module Fiddle
     # Fiddle::Importer#struct and Fiddle::Importer#union wrap this functionality in an
     # easy-to-use manner.
     #
-    # Example:
+    # Examples:
     #
     #   require 'fiddle/struct'
     #   require 'fiddle/cparser'
@@ -73,6 +73,17 @@ module Fiddle
     #
     #   MyStruct = Fiddle::CStructBuilder.create(Fiddle::CUnion, types, members)
     #
+    #   MyStruct.malloc(Fiddle::RUBY_FREE) do |obj|
+    #     ...
+    #   end
+    #
+    #   obj = MyStruct.malloc(Fiddle::RUBY_FREE)
+    #   begin
+    #     ...
+    #   ensure
+    #     obj.call_free
+    #   end
+    #
     #   obj = MyStruct.malloc
     #   begin
     #     ...
@@ -82,8 +93,8 @@ module Fiddle
     #
     def create(klass, types, members)
       new_class = Class.new(klass){
-        define_method(:initialize){|addr|
-          @entity = klass.entity_class.new(addr, types)
+        define_method(:initialize){|addr, func = nil|
+          @entity = klass.entity_class.new(addr, types, func)
           @entity.assign_names(members)
         }
         define_method(:[]) { |*args| @entity.send(:[], *args) }
@@ -100,9 +111,19 @@ module Fiddle
         def new_class.size()
           #{size}
         end
-        def new_class.malloc()
+        def new_class.malloc(func = nil)
           addr = Fiddle.malloc(#{size})
-          new(addr)
+          struct = new(addr, func)
+          if block_given?
+            raise ArgumentError, 'a free function must be supplied to malloc when it is called with a block' unless func
+            begin
+              yield struct
+            ensure
+              struct.to_ptr.call_free
+            end
+          else
+            struct
+          end
         end
       EOS
       return new_class
@@ -119,8 +140,18 @@ module Fiddle
     #
     # See Fiddle::Pointer.malloc for memory management issues.
     def CStructEntity.malloc(types, func = nil)
-      addr = Fiddle.malloc(CStructEntity.size(types))
-      CStructEntity.new(addr, types, func)
+      addr = Fiddle.malloc(self.size(types))
+      struct = new(addr, types, func)
+      if block_given?
+        raise ArgumentError, 'a free function must be supplied to Fiddle::CStructEntity.malloc when it is called with a block' unless func
+        begin
+          yield struct
+        ensure
+          struct.call_free
+        end
+      else
+        struct
+      end
     end
 
     # Returns the offset for the packed sizes for the given +types+.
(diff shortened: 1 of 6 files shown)


lib/fiddle/struct.rb
```diff
@@ -119,8 +140,18 @@ class CStructEntity < Fiddle::Pointer
     #
     # See Fiddle::Pointer.malloc for memory management issues.
     def CStructEntity.malloc(types, func = nil)
-      addr = Fiddle.malloc(CStructEntity.size(types))
-      CStructEntity.new(addr, types, func)
+      addr = Fiddle.malloc(self.size(types))
+      struct = new(addr, types, func)
```
kou, line 144: If an exception is raised in here, `addr` isn't freed.

chrisseaton: Yeah maybe I should use the pointer constructor rather than using `malloc` directly as it used to.

chrisseaton: Fixed.

test/fiddle/test_c_struct_entry.rb
```diff
@@ -43,37 +43,131 @@ def test_class_size_with_count
...
+    def test_new_double_free
+      types = [TYPE_INT]
+      Pointer.malloc(CStructEntity.size(types), Fiddle::RUBY_FREE) do |pointer|
+        assert_raise ArgumentError do
+          CStructEntity.new(pointer, types, Fiddle::RUBY_FREE)
+        end
+      end
+    end
+
+    def test_new_double_free
```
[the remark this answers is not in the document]
chrisseaton: Fixed.

chrisseaton:
This is the end of this series of PRs tweaking the API.

I do have some ideas about how to improve the implementation. I've got some goals to be able to share code between Fiddle and FFI, in order to make it easier for alternative implementations of Ruby which don't want to have to implement both interfaces. And I've got some goals to allow pointer operations to optimise much better in optimising implementations of Ruby.

lib/fiddle/struct.rb
```diff
@@ -94,33 +109,53 @@ def create(klass, types, members)
...
+            ensure
+              struct.to_ptr.call_free
+            end
+          else
+            struct
+          end
+        end
       }
       size = klass.entity_class.size(types)
+      # we use eval here make size into a literal, for performance
```
kou, line 127: We can use `define_singleton_method` for `size` here because this method will not be called too much.

lib/fiddle/struct.rb
```diff
@@ -94,33 +109,53 @@ def create(klass, types, members)
...
+      if block_given?
+        raise ArgumentError, 'a free function must be supplied to Fiddle::CStructEntity.malloc when it is called with a block' unless func
+        begin
+          yield struct
+        ensure
+          struct.call_free
+        end
+      else
+        struct
+      end
```
kou, line 158: How about reusing `malloc` block implementation?

If we reuse `malloc` block implementation, the allocated pointer is freed immediately when an exception is raised in `set_ctypes`.

```suggestion
      if block_given?
        super(size, func) do |struct|
          struct.set_ctypes types
          yield struct
        end
      else
        struct = super(size, func)
        struct.set_ctypes types
        struct
      end
```

lib/fiddle/struct.rb
```diff
@@ -152,6 +187,9 @@ def CStructEntity.size(types)
     #
     # See also Fiddle::Pointer.new
     def initialize(addr, types, func = nil)
+      if func && addr.is_a?(Pointer) && addr.free
+        raise ArgumentError, 'free function specified on both underlying struct Pointer and when creating a CStructEntity - who do you want to free this?'
+      end
```
kou, line 192: Why should we need to accept `Fiddle::Pointer` as `addr`?

lib/fiddle/struct.rb
```diff
@@ -261,21 +299,24 @@ def []=(*args)
...
     include PackInfo
 
     # Allocates a C union the +types+ provided.
     #
     # See Fiddle::Pointer.malloc for memory management issues.
     def CUnionEntity.malloc(types, func=nil)
-      addr = Fiddle.malloc(CUnionEntity.size(types))
-      CUnionEntity.new(addr, types, func)
+      super
     end
```
kou, line 320: Could you remove this method?

lib/fiddle/struct.rb
```diff
@@ -261,21 +299,24 @@ def []=(*args)
       end
     end
 
+    def size=(new_size) # :nodoc:
+      raise ArgumentError, 'cannot change the size of a struct'
+    end
+
```
kou, line 305: How about using `undef_method`?

```suggestion
    undef_method :size=
```

lib/fiddle/struct.rb
chrisseaton: Done.

lib/fiddle/struct.rb
chrisseaton: That's what `addr` is - a pointer - why wouldn't we accept a pointer where the user wants to give us an address? It's currently supported by the API - I didn't add it - I just made it safer here by not allowing two things to think they're responsible for the free operation.

lib/fiddle/struct.rb
chrisseaton: I didn't add it - but removed as you've asked.

lib/fiddle/struct.rb
kou: I see.

kou:
Thanks!

PR #42, opened 2020-06-21
mswin CI job doesn't work - remove it

chrisseaton:
Can we remove this failing job, and put it back in a PR where we get it working before merging it?

### DIFF
--- a/.github/workflows/fiddle.yml
+++ b/.github/workflows/fiddle.yml
@@ -14,7 +14,7 @@ jobs:
         ruby: [ 2.3, 2.4, 2.5, 2.6, 2.7, head ]
         include:
           - { os: windows , ruby: mingw }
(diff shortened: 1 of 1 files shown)


.github/workflows/fiddle.yml
```diff
@@ -14,7 +14,6 @@ jobs:
         ruby: [ 2.3, 2.4, 2.5, 2.6, 2.7, head ]
         include:
           - { os: windows , ruby: mingw }
-          - { os: windows , ruby: mswin }
```
kou, line 17: Could you comment it out with TODO comment to re-add easily in the future?

chrisseaton: Done.

kou:
Thanks!

PR #44, opened 2020-08-01
Add a "pinning" reference

tenderlove:
A `Fiddle::Pinned` objects will prevent the objects they point to from
moving.  This is useful in the case where you need to pass a reference
to a C extension that keeps the address in a global and needs the
address to be stable.

For example:

```ruby
class Foo
  A = "hi" # this is an embedded string

  some_c_function A # A might move!
end
```

If `A` moves, then the underlying string buffer may also move.
`Fiddle::Pinned` will prevent the object from moving:

```ruby
class Foo
  A = "hi" # this is an embedded string

  A_pinner = Fiddle::Pinned.new(A) # :nodoc:

  some_c_function A # A can't move because of `Fiddle::Pinned`
end
```

This is a similar strategy to what Graal uses:

  https://www.graalvm.org/sdk/javadoc/org/graalvm/nativeimage/PinnedObject.html#getObject--

### DIFF
--- a/ext/fiddle/pinned.c
+++ b/ext/fiddle/pinned.c
@@ -102,6 +102,8 @@ cleared_p(VALUE self)
     }
 }
 
+extern VALUE rb_eFiddleError;
+
 void
 Init_fiddle_pinned(void)
 {
@@ -117,5 +119,5 @@ Init_fiddle_pinned(void)
      *
      * Cleared reference exception
      */
-    rb_eFiddleClearedReferenceError = rb_define_class_under(mFiddle, "ClearedReferenceError", rb_eStandardError);
+    rb_eFiddleClearedReferenceError = rb_define_class_under(mFiddle, "ClearedReferenceError", rb_eFiddleError);
 }
--- a/ext/fiddle/pinned.c
+++ b/ext/fiddle/pinned.c
@@ -1,0 +1,121 @@
+#include <fiddle.h>
+
+VALUE rb_cPinned;
+VALUE rb_eFiddleClearedReferenceError;
+
+struct pinned_data {
+    VALUE ptr;
+};
+
+static void
+pinned_mark(void *ptr)
+{
+    struct pinned_data *data = (struct pinned_data*)ptr;
+    /* Ensure reference is pinned */
+    if (data->ptr) {
+        rb_gc_mark(data->ptr);
+    }
+}
+
+static size_t
+pinned_memsize(const void *ptr)
+{
+    return sizeof(struct pinned_data);
+}
+
+static const rb_data_type_t pinned_data_type = {
+    "fiddle/pinned",
+    {pinned_mark, xfree, pinned_memsize, },
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED
+};
+
+static VALUE
+allocate(VALUE klass)
+{
+    struct pinned_data *data;
+    VALUE obj = TypedData_Make_Struct(klass, struct pinned_data, &pinned_data_type, data);
+    data->ptr = 0;
+    return obj;
+}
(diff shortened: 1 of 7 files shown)


ext/fiddle/pinned.c
```diff
@@ -0,0 +1,121 @@
...
+    rb_define_method(rb_cPinned, "ref", ref, 0);
+    rb_define_method(rb_cPinned, "clear", clear, 0);
+    rb_define_method(rb_cPinned, "cleared?", cleared_p, 0);
+
+    /*
+     * Document-class: Fiddle::ClearedReferenceError
+     *
+     * Cleared reference exception
+     */
+    rb_eFiddleClearedReferenceError = rb_define_class_under(mFiddle, "ClearedReferenceError", rb_eStandardError);
```
kou, line 120: How about creating `class Fiddle::Error < StandardError; end` and use it as the parent error.
(We can use `Fiddle::Error` as the parent error of `Fiddle::DLError` too in the future.)

tenderlove: Ya, makes sense. I was not sure if I should do something like that in the same PR (but I will do it)

tenderlove: I added the class and also fixed `Fiddle::DLError`.

chrisseaton:
@tenderlove when you pin an object is the object pinned forever? I would have added an API like

```ruby
Fiddle::Pinned.new(A) do
  # pinned for the duration of the block
end
```

which would pin the object for the duration of some code and then un-pin it. Maybe if pinning is forever that's why you didn't add this though?

eregon:
This API seems overly broad to me, I think we only want to be able to pin Ruby Strings, and nothing else.

I believe it's impossible to pin arbitrary objects on HotSpot (the link above is about Native Image/SubstrateVM).
Some of the GCs just don't allow that as it makes significant trade-offs on compaction, efficiency, etc.

For the case of Strings OTOH that seems more reasonable, and that's trivial to implement on TruffleRuby because the `RSTRING_PTR(str)` never move on TruffleRuby (it' a `malloc()`).

ext/fiddle/pinned.c
```diff
@@ -0,0 +1,123 @@
...
+    VALUE ptr;
+};
+
+static void
+pinned_mark(void *ptr)
+{
+    struct pinned_data *data = (struct pinned_data*)ptr;
+    /* Ensure reference is pinned */
+    if (data->ptr) {
+        rb_gc_mark(data->ptr);
```
chrisseaton, line 16: Do you think we want a more explicit API in MRI to pin? You and I understand why this would pin the object, not sure everyone would.

tenderlove:
Sorry to take so long to get back to this.  @chrisseaton it only pins as long as the "pinning" object is alive or until the pinned reference is cleared.

Your example could be implemented like this:

```
require "fiddle"

def limited_pin object
  pin = Fiddle::Pinned.new object
  yield
  pin.clear
end

A = "foo"

limited_pin(A) do
  # whatever
end
```

> This API seems overly broad to me, I think we only want to be able to pin Ruby Strings' char pointer, and nothing else.

In the context of Fiddle or FFI extensions, I don't think this is too broad.  For example, some FFI extensions may want to save a Ruby object in to a `void *` and expect that the Ruby object's address will not change.  In that case we would need a pinning object such as this.

eregon:
> In the context of Fiddle or FFI extensions, I don't think this is too broad. For example, some FFI extensions may want to save a Ruby object in to a `void *` and expect that the Ruby object's address will not change. In that case we would need a pinning object such as this.

I see, makes sense.
I guess it's fine as a VALUE stored in native memory is a handle on TruffleRuby, and that never moves.
So probably we can just force creating the native handle for `Fiddle::Pinned.new object`.

PR #54, opened 2020-12-08
Add experimental MemoryView support

mrkn:


chrisseaton:
What is a `MemoryView`?

mrkn:
@chrisseaton See https://bugs.ruby-lang.org/issues/14722

mrkn:
I confirmed we can merge this pull-request to the latest master branch of ruby.
See https://github.com/ruby/ruby/pull/3860/checks?check_run_id=1520814844

chrisseaton:
Is it possible to provide more concrete use-cases in code or documentation? I'm keen to review and provide feedback, as I also want more direct memory buffers for various uses - for example reading binary files in my case. I can't really tell how I'd use this at the moment.

mrkn:
@chrisseaton MemoryView is designed to access a homogeneous array, such as a numeric tensor, placed on memory. I designed MemoryView by referring Python's BufferProtocol.  MemoryView in Ruby core is designed only for C extension libraries. `Fiddle::MemoryView` introduced in this pull-request is a wrapper object of that MemoryView.

I'll write documentation of MemoryView in Ruby core later, but you will able to understand how to use it by reading test codes of MemoryView in Ruby core and this pull-request.

dsisnero:
So, in order to use a memoryview from ruby and not c, we have to used fiddle/memoryview? 
Are IO objects going to be able to read to and from memoryview ?  
Apache::Arrow , Numo::NARRAY - are these going to have memoryview representations?

I am trying to wrap my head around how this relates to python buffer protocol and their python MemoryView

Thanks

mrkn:
@dsisnero An IO object is not a target of MemoryView.  MemoryView is designed to access a contiguous memory area.  If you want to access a file via MemoryView, you need to write a C extension library using mmap.

I'll write MemoryView supports of red-arrow and numo-narray asap.  After that, I'd like to let rmagick support MemoryView.

mrkn:
Note that MemoryView is an experimental feature in Ruby 3.0.0.
The API may be changed after releasing Ruby 3.0.0.

dsisnero:
from fiddle/memoryview can we slice and return another memoryview object?  
 str = Marshal.load(Marshal.dump("hello world"))
 ptr = Pointer[str]
 mview = MemoryView.new(ptr)
tail = mview[1..-1]      # return new MemoryView object of  mview slice

This shows what I was thinking we could do with Fiddle/memoryview

https://julien.danjou.info/high-performance-in-python-with-zero-copy-and-the-buffer-protocol/

mrkn:
@dsisnero I don't include the sub-memoryview slicing feature in this pull-request.  But, I want to support it in the future version.

PR #55, opened 2020-12-08
Use libSystem.B.dylib instead of libm.dylib and libc.dylib

mrkn:
Because macOS 11.0(Big Sur) was removed libc and libm from `/usr/lib`.

~https://github.com/ruby/ruby/commit/bd47a8d660~
https://github.com/ruby/ruby/commit/b7d86e330c
https://github.com/ruby/ruby/commit/cdef17096c
https://github.com/ruby/ruby/commit/0168094da2

### DIFF
--- a/test/fiddle/helper.rb
+++ b/test/fiddle/helper.rb
@@ -55,8 +55,7 @@ when /mingw/, /mswin/
   crtname = RbConfig::CONFIG["RUBY_SO_NAME"][/msvc\w+/] || 'ucrtbase'
   libc_so = libm_so = "#{crtname}.dll"
 when /darwin/
-  libc_so = "libc.dylib"
-  libm_so = "libm.dylib"
+  libc_so = libm_so = "/usr/lib/libSystem.B.dylib"
 when /kfreebsd/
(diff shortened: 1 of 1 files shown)


kou:
`libc.dylib` and `libm.dylib` are removed but `dlopen("libc.dylib")` and `dlopen("libm.dylib")` are still work because `dlopen()` handles them specially.
I forgot URL that describes it...

kou:
Note that `libc.dylib` not `/usr/lib/libc.dylib` is important.

mrkn:
@kou I got DLError by `libc.dylib` on https://github.com/ruby/ruby/pull/3860/checks?check_run_id=1514983494

mrkn:
Maybe we should treat different darwin versions in different ways...

kou:
It's strange...
We have macOS 10.15 and 11.0 in ruby/fiddle and they work well with `libc.dylib`.

mrkn:
How do you think about using `/usr/lib/libSystem.B.dylib`?

kou:
It's OK to me if it works.

mrkn:
@kou thanks!

PR #60, opened 2020-12-23
Import from ruby

mrkn:


mrkn:
I ignore head, mswin, and mingw jobs because they are old.

chrisseaton:
Can you note which version of Ruby we've imported from here?

Ideally 3.0 would be on a release of FFI, not some in-between version as gems often end up being. This causes pretty severe confusion for downstream Ruby implementations and in some cases security problems.

mrkn:
The commits in this PR came from the master branch of ruby.
I need to sync both gem and ruby-bundled fiddle.
I'll release new version after emerging this branch.

chrisseaton:
Hmm yeah thanks for doing that! Not sure it's ideal that Ruby potentially release 3.0 with a version of FFI that doesn't match any release.

mrkn:
> Not sure it's ideal that Ruby potentially release 3.0 with a version of FFI that doesn't match any release.

What do you mean by this sentence?

The new version of fiddle can be used not only in ruby 3.0, but also in ruby 2.7, 2.6, and 2.5 (tests are passed on all these versions).  I don't think releasing new version is problematic.

mrkn:
Note that the latest release version of fiddle is already synced to old ruby's master branch.
And, This pull-request consists of the changes only related to MemoryView.

chrisseaton:
Sorry I thought these were changes that were already in RC1 and RC1 was claiming to be 1.0.4 but they didn't have the same content. This has happened on other gems and it's some nasty caused security confusion.

I checked the diffs and that's not the case though.

Sorry for wasting your time.

mrkn:
No problem 😄

PR #61, opened 2020-12-23
Refactored parse_signature

kojix2:
This is just an idea.
A proposal to improve the parse_signature method.

Background:
I wanted to parse_signature recursively when the argument was a function (i.e. a function pointer).
When I wrote a monkey patch for it, I noticed that parse_signature was unnecessarily complicated.

1. Line 114 and line 117 are almost the same code.
2. Each line is unnecessarily long.
3. Some variable names are confusing.
4. Also, the variables $1, $2, etc. looks old-fashioned.

2 - 4 are more of a library-wide issue, and are my personal opinions rather than objective facts.
Libraries have their own history.

Here, I am focusing on 1.

Please feel free to close this.
Thank you.

### DIFF
--- a/lib/fiddle/cparser.rb
+++ b/lib/fiddle/cparser.rb
@@ -108,16 +108,15 @@ module Fiddle
     #
     def parse_signature(signature, tymap=nil)
       tymap ||= {}
-      case compact(signature)
-      when /^(?:[\w\*\s]+)\(\*(\w+)\((.*?)\)\)(?:\[\w*\]|\(.*?\));?$/
-        func, args = $1, $2
-        return [func, TYPE_VOIDP, split_arguments(args).collect {|arg| parse_ctype(arg, tymap)}]
-      when /^([\w\*\s]+[\*\s])(\w+)\((.*?)\);?$/
-        ret, func, args = $1.strip, $2, $3
-        return [func, parse_ctype(ret, tymap), split_arguments(args).collect {|arg| parse_ctype(arg, tymap)}]
-      else
-        raise(RuntimeError,"can't parse the function prototype: #{signature}")
-      end
+      ret, func, args = case compact(signature)
+                        when /^(?:[\w*\s]+)\(\*(\w+)\((.*?)\)\)(?:\[\w*\]|\(.*?\));?$/
+                          [TYPE_VOIDP, $1, $2]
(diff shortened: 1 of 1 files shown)


kojix2:
Hey, wait a minute. I think I made a mistake.

Maumagnaguagno:
You modified the regular expressions, from `\*` to `*`.
Also, you are calling an extra ``parse_ctype(ret, tymap)`` instead of just ``TYPE_VOIDP``.
Probably better to keep such call inside the ``case`` statement: ``[parse_ctype($1.strip, tymap), $2, $3]``.

kojix2:
Fixed. 
@Maumagnaguagno I appreciate the help.

Maumagnaguagno:
I guess you need to propagate ``tymap`` to ``parse_ctype`` to avoid creating a new Hash in ``tymap ||= {}``.

kojix2:
Yes, I think you're right.
But this is the first time I've sent a PR to Fiddle.
I think I'll leave it as is and see what the project owner has to say about it...

Anyway, thank you again.

kojix2:
I went back to work on the monkey patch one more time, but it wouldn't work without passing the tymap  :facepalm: 
I have to fix it...

kou:
>  I wanted to parse_signature recursively when the argument was a function (i.e. a function pointer).

Could you show a signature you want to parse?

kou:
An example: `void uiSpinboxOnChanged(uiSpinbox *s, void (*f)(uiSpinbox *s, void *data), void *data)`

kojix2:
Yes, thanks.
I would like to create a customized method. When you pass a block to that method, it will create a `Fidle::Closure::BlockCaller` with the appropriate type. To do this, each function needs to know the type information beforehand.

kou:
Idea:

We use `[TYPE, SIZE]` for sized array type. e.g.: `[Fiddle::TYPE_CHAR, 80]` for `char x[80]`

It may be a good idea that we use `[Fiddle::TYPE_VOIDP, RETURN_TYPE, [ARGUMENT_TYPE, ...]]` for function pointer. e.g.: `[Fiddle::TYPE_VOIDP, Fiddle::TYPE_VOID, [Fiddle::TYPE_VOIDP, Fiddle::Type_VOIDP]]` for `void (*f)(uiSpinbox *s, void *data)`

PR #69, opened 2021-03-03
Use test-unit gem

hsbt:
I replaced the test-framework to test-unit instead of original framework of ruby/ruby. 

It's easy to maintain rather than the current test-suite.


hsbt:
@kou @mrkn How about this? I hope to merge and reduce code from deep copy from ruby repo.

kou:
I'll add `assert_*` methods that are used by Fiddle to test-unit and release test-unit.
Please wait for a while.

hsbt:
👍  Thanks. I'll migrate it and using `omit` after that.

kou:
I'll do them too.
You can just wait for my work. :-)

hsbt:
> I'll do them too.

Ah. I means that migration is without fiddle repo like rdoc, irb and etc.

kou:
Ah, I see.

hsbt:
👍

kou:
Done.

PR #70, opened 2021-03-10
Do not use a libdir for glibc, it breaks Linux PPC64

jeremyevans:
Fixes [Bug #12666]

### DIFF
--- a/test/fiddle/helper.rb
+++ b/test/fiddle/helper.rb
@@ -47,8 +47,8 @@ when /linux/
     libm_so = libc_so
   else
     # glibc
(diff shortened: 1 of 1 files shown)


kou:
Thanks!

ahorek:
you may consider using the same trick as FFI
https://github.com/ffi/ffi/blob/master/lib/ffi/platform.rb#L141
https://github.com/ffi/ffi/blob/master/ext/ffi_c/Platform.c#L42

otherwise, ```libc.so.6``` without an absolute path should be preferred.

PR #80, opened 2021-06-16
Add MemoryView.export and MemoryView#release

### LINKED ISSUE #79: Segfault when exiting due to `Fiddle::MemoryView`
The following script causes a segfault when exiting:

```
$workaround = false

require 'fiddle'

Fiddle::MemoryView.new(Fiddle::Pointer["foo"])

GC.start if $workaround
```

A backtrace from GDB:

```
Thread 1 received signal SIGSEGV, Segmentation fault.
rb_class_get_superclass (klass=klass@entry=0) at ../object.c:2252
2252        return RCLASS(klass)->super;
(gdb) bt
#0  rb_class_get_superclass (klass=klass@entry=0) at ../object.c:2252
#1  0x0000000589c03ae3 in lookup_memory_view_entry (klass=0) at ../memory_view.c:784
#2  0x0000000589c04921 in rb_memory_view_release (view=view@entry=0x800157770) at ../memory_view.c:836
#3  0x0000000596513c13 in fiddle_memview_free (ptr=0x800157770) at ../../../ext/fiddle/memory_view.c:41
#4  0x0000000589bd9226 in run_final (zombie=123145300985120, objspace=0x800052bc0) at ../gc.c:4040
#5  finalize_list (objspace=objspace@entry=0x800052bc0, zombie=123145300985120) at ../gc.c:4059
#6  0x0000000589bdd6e1 in rb_objspace_call_finalizer (objspace=0x800052bc0) at ../gc.c:4222
#7  0x0000000589bc5f37 in rb_ec_finalize (ec=ec@entry=0x800053520) at ../eval.c:187
#8  0x0000000589bc8e25 in rb_ec_cleanup (ec=ec@entry=0x800053520, ex=<optimized out>) at ../eval.c:298
#9  0x0000000589bc8fa8 in ruby_run_node (n=0x6fffffed3068) at ../eval.c:379
#10 0x00000001004017ad in main (argc=<optimized out>, argv=<optimized out>) at ../main.c:47
```

Invoking a GC before exiting seems to be a workaround.

`ruby -v`: `ruby 3.1.0dev (2021-05-21T09:28:24Z master 50a534a152) [x86_64-linux]`

kou:
fix #79

Users can release memory views explicitly before process exit.

Reported by xtkoba. Thanks!!!

### DIFF
--- a/ext/fiddle/memory_view.c
+++ b/ext/fiddle/memory_view.c
@@ -94,7 +94,7 @@ rb_fiddle_memview_release(VALUE obj)
 }
 
 static VALUE
-rb_fiddle_memview_s_get(VALUE klass, VALUE target)
+rb_fiddle_memview_s_export(VALUE klass, VALUE target)
 {
     ID id_new;
     CONST_ID(id_new, "new");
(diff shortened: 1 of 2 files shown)


mrkn:
I think these two methods are useful.  But, it seems less clear that `get` with a block implicitly releases the memory view at the end of the block.  I want to change the name of `get` if we can find a more appropriate name.

kou:
`get` is borrowed from `rb_memory_view_get_func_t`.

Other candidates:

  * `MemoryView.open`: like `File.open` but `MemoryView` doesn't open anything
  * `MemoryView.acquire`: `release` and `acquire` pair is often used in "lock" context but the current context isn't "lock"
  * `MemoryView.map`
  * `MemoryView.view`

kou:
We choose `MemoryView.export`.

PR #83, opened 2021-06-30
Add "offsetof" to Struct classes

tenderlove:
I need to get the offset of a member inside a struct without allocating
the struct.  This patch adds an "offsetof" class method to structs that
are generated.

The usage is like this:

```ruby
MyStruct = struct [
  "int64_t i",
  "char c",
]

MyStruct.offsetof("i") # => 0
MyStruct.offsetof("c") # => 8
```

### DIFF
--- a/test/fiddle/test_c_struct_builder.rb
+++ b/test/fiddle/test_c_struct_builder.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+begin
+  require_relative 'helper'
+  require 'fiddle/struct'
+  require 'fiddle/cparser'
+rescue LoadError
+end
+
+module Fiddle
+  class TestCStructBuilder < TestCase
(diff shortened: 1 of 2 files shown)


test/fiddle/test_c_struct_builder.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    def test_union_offsetof
+      types, members = parse_struct_signature(['int64_t i','char c'])
+      my_struct = Fiddle::CStructBuilder.create(Fiddle::CUnion, types, members)
+      assert_equal 0, my_struct.offsetof("i")
+      assert_equal 0, my_struct.offsetof("c")
+    end
+  end
+end
```
kou, line 36:
```suggestion
end if defined?(Fiddle)
```

kou (approved):

PR #86, opened 2021-07-13
Check HAVE_RUBY_MEMORY_VIEW_H rather than API version

nobu:
These macros are defined for exactly this purpose.

### DIFF
--- a/ext/fiddle/fiddle.h
+++ b/ext/fiddle/fiddle.h
@@ -189,10 +189,6 @@
 #define ALIGN_INT32_T ALIGN_OF(int32_t)
 #define ALIGN_INT64_T ALIGN_OF(int64_t)
 
-#ifdef HAVE_RUBY_MEMORY_VIEW_H
-# define FIDDLE_MEMORY_VIEW
-#endif
-
 extern VALUE mFiddle;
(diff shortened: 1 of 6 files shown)


ext/fiddle/fiddle.h
```diff
@@ -190,7 +189,7 @@
 #define ALIGN_INT32_T ALIGN_OF(int32_t)
 #define ALIGN_INT64_T ALIGN_OF(int64_t)
 
-#if RUBY_API_VERSION_MAJOR >= 3
+#ifdef HAVE_RUBY_MEMORY_VIEW_H
 # define FIDDLE_MEMORY_VIEW
```
kou, line 193: Can we remove `FIDDLE_MEMORY_VIEW` and use `HAVE_RUBY_MEMORY_VIEW_H` directly?

nobu: Yes, unless you want to add other conditions.

kou: Could you also do it in this pull request?

nobu:
Seems to fmail when the latest (with debug or on Windows).

kou:
Thanks!

PR #87, opened 2021-07-13
Return the module handle value in Fiddle::Handle#to_i

mrkn:
`Fiddle::Handle#to_i` currently returns the address of the value of the module handle, that is `&fiddle_handle->ptr`.
I guess it is more useful than now that it returns the value of the module handle directly.

### DIFF
--- a/ext/fiddle/fiddle.h
+++ b/ext/fiddle/fiddle.h
@@ -199,5 +199,8 @@ extern VALUE rb_eFiddleDLError;
 
 VALUE rb_fiddle_new_function(VALUE address, VALUE arg_types, VALUE ret_type);
 
+typedef void (*rb_fiddle_freefunc_t)(void*);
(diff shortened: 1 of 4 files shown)


mrkn:
I get the another proposal from @unak to introduce the new method `Fiddle::Handle#value` for this purpose.

mrkn:
Other approach: Introducing `Fiddle::Handle#value` or `Fiddle::Handle#pointer` and let them return a `Fiddle::Pointer` instead of an integer.

kou:
This is a backward incompatible change but the current `Handle#to_i` is useless. So nobody will get a trouble.
I'm OK with this change.

I also like `Fiddle::Handle#pointer` approach but how about `#to_pointer` instead of `#pointer`?

mrkn:
OK. I will add `to_ptr` instead of `to_pointer` because `Fiddle::Pointer.[]` examine `to_ptr`.

kou:
Thanks!

PR #89, opened 2021-07-14
version.rb - increment for ruby master builds

MSP-Greg:
Ruby master builds have issues with installing an (extension?) gem when it is a default gem and the version matches the installed version.

Increment the version so master builds pass.

### DIFF
--- a/lib/fiddle/version.rb
+++ b/lib/fiddle/version.rb
@@ -1,3 +1,3 @@
 module Fiddle
-  VERSION = "1.1.0"
+  VERSION = "1.1.1"
 end


kou:
Wow! Thanks for investing this.

But we need to bump version again when we merge changes in ruby/fiddle into ruby/ruby. I've added a step that bumps version in CI.

MSP-Greg:
> But we need to bump version again with this approach when we merge changes in ruby/fiddle into ruby/ruby.

Just to be clear, I don't think it's an issue in ruby/ruby, since the testing is done against the default gem.  The issue is when the repo has the same version number as the Ruby master builds used in CI.  I think.

kou:
I think that I understand your explanation correctly.

If we merge this approach and this change is merged into ruby/ruby like https://github.com/ruby/ruby/commit/a2c9e1b58a9bc1da0533171236da43fcb556ead7, we need to bump our version again.

MSP-Greg:
Sorry AFK for a bit.  I think what you changed in CI (`rake version:bump`) is the best solution, but maybe do it for all jobs?  I'm thinking right after a new Ruby release, the versions could match?

kou:
Ah, I understand.
Generally, Ruby releases should not ship unreleased Fiddle. Version in ruby/fiddle will be bumped when we release a new Fiddle. So it should not be happen. The Ruby master may only include unreleased Fiddle.

PR #90, opened 2021-08-19
Improve "offsetof" calculations

tenderlove:
I need to get the offset of members inside sub structures.  This patch
adds sub-structure offset support for structs.


kou (approved):
+1

PR #94, opened 2021-09-24
Add CI steps for `--enable-bundled-libffi`

tenderlove:
@hsbt reported an issue with `--enable-bundled-libffi` [here](https://bugs.ruby-lang.org/issues/18034).

I added a build that uses this flag, and that build is failing.  I was able to reproduce @hsbt's issue.  I think it's caused by libffi installed with homebrew *and* using the `--enable-bundled-libffi` flag.  My guess is that building the bundled libffi is failing, but the gem still compiles with the bundled ffi headers but links against the homebrew libffi.

I'm not sure how to fix yet, but I wanted to send this PR to show the issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        build_flags:
+          - ""
+          - " -- --enable-bundled-libffi"
         os:
           - macos-10.15
           - macos-11.0
@@ -40,7 +43,7 @@ jobs:
 
       - run: bundle install
 
-      - run: rake compile
+      - run: rake compile ${{ matrix.build_flags }}
 
       # If Fiddle in Ruby's master has the same version of this Fiddle,
       # "gem install pkg/*.gem" fails.


.github/workflows/ci.yml
```diff
@@ -12,6 +12,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        build_flags:
+          - ""
+          - " -- --enable-bundled-libffi"
```
kou, line 17: Could you add this information to `name`?
We can't distinct this information from the current job list...

* macos-10.15 debug: https://github.com/ruby/fiddle/pull/94/checks?check_run_id=3701967650
* macos-10.15 debug: https://github.com/ruby/fiddle/pull/94/checks?check_run_id=3701968329

kou (commented):
It seems that SEGV isn't reproduced in CI. The failed job is just failed by `extconf.rb`: https://github.com/ruby/fiddle/pull/94/checks?check_run_id=3701968329
How about installing libffi by Homebrew when `--enable-bundled-libffi` is used in the job?

hsbt:
I removed `--enabled-bundled-libffi` feature at https://github.com/ruby/fiddle/pull/113

PR #98, opened 2021-11-30
Enabled `--enabled-bundled-ffi` option for only mswin environment

hsbt:
We discussed at https://bugs.ruby-lang.org/issues/18034

### DIFF
--- a/ext/fiddle/extconf.rb
+++ b/ext/fiddle/extconf.rb
@@ -46,7 +46,7 @@ end
 
 libffi_version = nil
 have_libffi = false
-bundle = enable_config('bundled-libffi')
+bundle = enable_config('bundled-libffi') && $mswin
 unless bundle
   dir_config 'libffi'
 


ext/fiddle/extconf.rb
```diff
@@ -46,7 +46,7 @@ def enable_debug_build_flag(flags)
 
 libffi_version = nil
 have_libffi = false
-bundle = enable_config('bundled-libffi')
+bundle = enable_config('bundled-libffi') && $mswin
```
kou, line 49: Can we raise an exception if an user specifies `--enable-bundled-ffi` on non mswin environment?

larskanis:
ffi.gem bundles libffi since more than 10 years. The extconf.rb file checks whether system libffi is usable and otherwise falls back to builtin. Since commit https://github.com/ffi/ffi/commit/f0959778cfac41e4673117993c3abf26ef1ae76c it's also possible to force system or builtin libffi, so that users with an issue with the default selection can switch to the other way.

libffi on Macos is a mess. We had several bug reports which showed, that that header files don't fit to the libraries or that very old libffi versions were installed as system libffi. We did some fine tuning and now the reports are quite rare. Or maybe newer Macos versions fixed these issues or just stackoverflow has enough answers in the meantime.

Nevertheless builtin libffi works just fine on most operation systems and we test this on Windows, Linux and Macos: https://github.com/ffi/ffi/actions/runs/1287629846

hsbt:
I and @unak try to use libffi provided by vcpkg. After that, We can remove the bundled source of libffi for mswin environment.

kou:
@larskanis @hsbt Thanks for the information.
I'll consider libffi bundling again in this month.

hsbt:
bundled source has been removed at https://github.com/ruby/ruby/commit/e4f5296f065110fa83eb450d3a861253e76e534f

kou:
Let's remove bundled libffi in ruby/fiddle too.

hsbt:
I'm going to submit another PR for removing download feature.

PR #101, opened 2021-12-19
Update CParser#parse_struct_signature to support "int long blah"

### LINKED ISSUE #100: parse_ctype() fails for "int long len"
`parse_ctype()` does not handle `int long len` correctly; errors with `unknown type: int`

```ruby
[22] pry(#<Fiddle::CastXMLImporter::Importer>)> parse_struct_signature(["char id", "int long size"])
Fiddle::DLError: unknown type: int
from /Users/dmlary/.rbenv/versions/3.0.2/lib/ruby/3.0.0/fiddle/cparser.rb:245:in `parse_ctype'
[24] pry(#<Fiddle::CastXMLImporter::Importer>)> parse_ctype("int long size")
Fiddle::DLError: unknown type: int
from /Users/dmlary/.rbenv/versions/3.0.2/lib/ruby/3.0.0/fiddle/cparser.rb:245:in `parse_ctype'
[25] pry(#<Fiddle::CastXMLImporter::Importer>)> parse_ctype("int long")
=> 4
```

I am aware that I can switch the field definition to `long`, but these types are being pulled out of CastXML, and I'd prefer to fix the bug in Fiddle instead of implementing a bunch of type remapping.

### old
~~`parse_struct_signature()` is passing the complete struct member definition to `parse_ctype()`.  This includes both the ctype and the member name.  This causes a problem when handling structs that contain compound members such as: `int long length` in their definition.~~

dmlary:
fixes #100 

(was hoping this would run the pipeline, but no luck; I'll find a way to run the tests)

### DIFF
--- a/lib/fiddle/cparser.rb
+++ b/lib/fiddle/cparser.rb
@@ -66,12 +66,12 @@ module Fiddle
             mems.push([struct_name, struct_type.members])
             tys.push(ty)
           end
-        when /^[\w\*\s]+[\*\s](\w+)$/
-          mems.push($1)
-          tys.push(parse_ctype(msig, tymap))
-        when /^[\w\*\s]+\(\*(\w+)\)\(.*?\)$/
-          mems.push($1)
-          tys.push(parse_ctype(msig, tymap))
+        when /^([\w\*\s]+[\*\s])(\w+)$/
+          mems.push($2)
+          tys.push(parse_ctype($1, tymap))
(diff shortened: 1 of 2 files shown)


dmlary:
A complete fix for `parse_ctype()` is beyond what I would want to do for a first commit to this project.  See https://github.com/ruby/fiddle/issues/100#issuecomment-997486485 for details

PR #108, opened 2022-05-04
Add `sym_defined?` methods to test if a symbol is defined

tenderlove:
I would like to check if a symbol is defined before trying to access it.  Some symbols aren't available on all platforms, so instead of raising an exception, I want to check if it's defined first.

Today we have to do:

```ruby
begin
  addr = Fiddle::Handle.sym("something")
  # do something
rescue Fiddle::DLError
end
```

I want to write this:

```ruby
if Fiddle::Handle.sym_defined?("something")
  addr = Fiddle::Handle.sym("something")
  # do something
end
```

### DIFF
--- a/ext/fiddle/handle.c
+++ b/ext/fiddle/handle.c
@@ -321,8 +321,8 @@ rb_fiddle_handle_s_sym(VALUE self, VALUE sym)
     return fiddle_handle_sym(RTLD_NEXT, sym);
 }
 
-static VALUE
-fiddle_handle_sym(void *handle, VALUE symbol)
+static void *
+fiddle_handle_find_func(void *handle, VALUE symbol)
 {
 #if defined(HAVE_DLERROR)
     const char *err;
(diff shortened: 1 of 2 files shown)


ext/fiddle/handle.c
```diff
@@ -321,8 +321,8 @@ rb_fiddle_handle_s_sym(VALUE self, VALUE sym)
     return fiddle_handle_sym(RTLD_NEXT, sym);
 }
 
-static VALUE
-fiddle_handle_sym(void *handle, VALUE symbol)
+static void *
```
kou, line 324: Should we use `void (*)()` as a return type instead of `void *`?
But we can't write it as `static (void (*)()) fiddle_handle_find_func(...) {...}`. We need to write it as `static void (*fiddle_handle_find_func(void *handle, VALUE symbol)) () {...}`. But it's difficult to read...

It's better that we define a type for it like:

```diff
diff --git a/ext/fiddle/handle.c b/ext/fiddle/handle.c
index 652cf77..c8a7a0a 100644
--- a/ext/fiddle/handle.c
+++ b/ext/fiddle/handle.c
@@ -321,7 +321,9 @@ rb_fiddle_handle_s_sym(VALUE self, VALUE sym)
     return fiddle_handle_sym(RTLD_NEXT, sym);
 }
 
-static void *
+typedef void (*fiddle_void_func)(void);
+
+static fiddle_void_func
 fiddle_handle_find_func(void *handle, VALUE symbol)
 {
 #if defined(HAVE_DLERROR)
@@ -330,13 +332,13 @@ fiddle_handle_find_func(void *handle, VALUE symbol)
 #else
 # define CHECK_DLERROR
 #endif
-    void (*func)();
+    fiddle_void_func func;
     const char *name = StringValueCStr(symbol);
 
 #ifdef HAVE_DLERROR
     dlerror();
 #endif
-    func = (void (*)())(VALUE)dlsym(handle, name);
+    func = (fiddle_void_func)(VALUE)dlsym(handle, name);
     CHECK_DLERROR;
 #if defined(FUNC_STDCALL)
     if( !func ){
@@ -386,7 +388,7 @@ fiddle_handle_find_func(void *handle, VALUE symbol)
 static VALUE
 rb_fiddle_handle_s_sym_defined(VALUE self, VALUE sym)
 {
-    void (*func)();
+    fiddle_void_func func;
 
     func = fiddle_handle_find_func(RTLD_NEXT, sym);
...
```

BTW, it seems that we don't call the `func`. How about we just use `void *` instead of `void (*func)()`?

chrisseaton: > How about we just use `void *` instead of `void (*func)()`?

Isn't that what it currently does?

There's also no information about the function's return type - so why would we assert in the type that it returns `void`?

I don't think you can do anything useful or intelligent with these types.

kou: Ah, I should have said it more explicitly.
I meant it for `func` local variables not the return type of this function.

tenderlove: I agree, it makes the code much easier to read.  I added the change!

ext/fiddle/handle.c
```diff
@@ -379,6 +381,53 @@ fiddle_handle_sym(void *handle, VALUE symbol)
...
+    else {
+	return Qnil;
+    }
+}
+
+static VALUE
+rb_fiddle_handle_sym_defined(VALUE self, VALUE sym)
+{
+    struct dl_handle *fiddle_handle;
+    void (*func)();
```
kou, line 407:
```suggestion
    fiddle_void_func func;
```

ext/fiddle/handle.c
```diff
@@ -379,6 +381,53 @@ fiddle_handle_sym(void *handle, VALUE symbol)
...
+    }
+    else {
+	return Qnil;
+    }
+}
+
+static VALUE
+fiddle_handle_sym(void *handle, VALUE symbol)
+{
+    void (*func)();
```
kou, line 427:
```suggestion
    fiddle_void_func func;
```

kou (approved):
+1

PR #110, opened 2022-09-05
Fix PACK_MAP for unsigned types

k0kubun:
I think void* and unsigned types should be packed/unpacked as unsigned values.


lib/fiddle/pack.rb
```diff
@@ -48,7 +48,7 @@ module PackInfo # :nodoc: all
       ALIGN_MAP[TYPE_LONG_LONG] = ALIGN_MAP[-TYPE_LONG_LONG] = ALIGN_LONG_LONG
       PACK_MAP[TYPE_LONG_LONG] = PACK_MAP[-TYPE_LONG_LONG] = "q"
```
kou, line 49: Should we use `Q` for `-TYPE_LONG_LONG` too?

k0kubun: Good catch, thank you https://github.com/ruby/fiddle/pull/110/commits/43b1fe4cfd95b719d8bef67f3b49d60f439cfd9e

kou (approved):
+1

k0kubun:
FYI, I cherry-picked this to ruby/ruby https://github.com/ruby/ruby/commit/106744107b485a371ed788fe590c6ce778de16df as I needed to use this in ruby/ruby (MJIT).

I'd also like to cherry-pick https://github.com/ruby/fiddle/pull/111 if you're okay with it.

kou:
I'm OK with it.
You can merge the current master instead of cherry-picking some commits if you want.

k0kubun:
got it, thank you!

PR #111, opened 2022-09-10
Add constants for unsigned values

tenderlove:
This commit adds constants for unsigned values.  Currently we can use `-` to mean "unsigned", but I think having a specific name makes Fiddle more user friendly.  This commit continues to support `-`, but introduces negative constants with "unsigned" names

I think this will help to eliminate [this code](https://github.com/ruby/ruby/blob/3a56bf0bcc66e14ffe5ec89efc32ecfceed180f4/lib/mjit/c_type.rb#L31-L38)

### DIFF
--- a/lib/fiddle/cparser.rb
+++ b/lib/fiddle/cparser.rb
@@ -164,23 +164,23 @@ module Fiddle
         unless Fiddle.const_defined?(:TYPE_LONG_LONG)
           raise(RuntimeError, "unsupported type: #{ty}")
         end
-        return -TYPE_LONG_LONG
+        return TYPE_ULONG_LONG
       when /\A(?:signed\s+)?long(?:\s+int\s+)?(?:\s+\w+)?\z/
         return TYPE_LONG
       when /\Aunsigned\s+long(?:\s+int\s+)?(?:\s+\w+)?\z/
-        return -TYPE_LONG
+        return TYPE_ULONG
       when /\A(?:signed\s+)?int(?:\s+\w+)?\z/
         return TYPE_INT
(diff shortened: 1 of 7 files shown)


k0kubun (approved):
I'd be happy to have these constants 🙂

lib/fiddle/cparser.rb
```diff
@@ -210,7 +210,7 @@ def parse_ctype(ty, tymap=nil)
         unless Fiddle.const_defined?(:TYPE_INT32_T)
           raise(RuntimeError, "unsupported type: #{ty}")
         end
-        return -TYPE_INT32_T
+        return TYPE_INT32_T
```
kou, line 213:
```suggestion
        return TYPE_UINT32_T
```

test/fiddle/test_import.rb
```diff
@@ -130,6 +130,8 @@ def test_sizeof()
         name = $1.sub(/P\z/,"*").gsub(/_(?!T\z)/, " ").downcase
         type_name = name
       end
+      type_name = "unsigned #{$1}" if type_name =~ /u(long|short|char)/
```
kou, line 133:
```suggestion
      type_name = "unsigned #{$1}" if type_name =~ /\Au(long|short|char)/
```

Do we need to add `int` too here?

kou (approved):
+1

test/fiddle/test_import.rb
tenderlove: We might have a missing test.  It seems like we need `int`, but no tests fail.  I will investigate

PR #112, opened 2022-09-12
Move "type" constants to `Fiddle::Types`

tenderlove:
This helps to reduce repetition in code.  Instead of doing "TYPE_*" everywhere, you can do `include Fiddle::Types`, and write the type name directly.

This PR is to help reduce repetition when writing Fiddle code.  Right now we have to type `TYPE_` everywhere, and you also have to include all of `Fiddle` to access `TYPE_*` constants.  With this change, you can just include `Fiddle::Types` and it will shorten your code and also you only have to include those constants.

Here is an example before:

```ruby
require "fiddle"

module MMAP
  # All Fiddle constants included
  include Fiddle

  def self.make_function name, args, ret
    ptr = Handle::DEFAULT[name]
    func = Function.new ptr, args, ret, name: name
    define_singleton_method name, &func.to_proc
  end

  make_function "munmap", [TYPE_VOIDP, # addr
                           TYPE_SIZE_T], # len
                           TYPE_INT

  make_function "mmap", [TYPE_VOIDP,
                         TYPE_SIZE_T,
                         TYPE_INT,
                         TYPE_INT,
                         TYPE_INT,
                         TYPE_INT], TYPE_VOIDP

  make_function "mprotect", [TYPE_VOIDP, TYPE_SIZE_T, TYPE_INT], TYPE_INT
end
```

After:

```ruby
require "fiddle"

module MMAP
  # Only type names included
  include Fiddle::Types

  def self.make_function name, args, ret
    ptr = Fiddle::Handle::DEFAULT[name]
    func = Fiddle::Function.new ptr, args, ret, name: name
    define_singleton_method name, &func.to_proc
  end

  make_function "munmap", [VOIDP, # addr
                           SIZE_T], # len
                           INT

  make_function "mmap", [VOIDP, SIZE_T, INT, INT, INT, INT], VOIDP

  make_function "mprotect", [VOIDP, SIZE_T, INT], INT
end
```

We only need to import the type names, and you don't have to type `TYPE_` over and over.  I think this makes Fiddle code easier to read.

### DIFF
--- a/ext/fiddle/fiddle.c
+++ b/ext/fiddle/fiddle.c
@@ -346,7 +346,7 @@ Init_fiddle(void)
      */
     rb_define_const(mFiddleTypes, "INTPTR_T",  INT2NUM(TYPE_INTPTR_T));
 
-    /* Document-const: Fiddle::TypesUINTPTR_T
+    /* Document-const: Fiddle::Types::UINTPTR_T
      *
      * C type - uintptr_t
      */
--- a/ext/fiddle/fiddle.c
+++ b/ext/fiddle/fiddle.c
@@ -164,191 +164,193 @@ Init_fiddle(void)
      */
     rb_eFiddleDLError = rb_define_class_under(mFiddle, "DLError", rb_eFiddleError);
 
-    /* Document-const: TYPE_VOID
+    VALUE mFiddleTypes = rb_define_module_under(mFiddle, "Types");
+
+    /* Document-const: Fiddle::Types::VOID
      *
      * C type - void
      */
-    rb_define_const(mFiddle, "TYPE_VOID",      INT2NUM(TYPE_VOID));
+    rb_define_const(mFiddleTypes, "VOID",      INT2NUM(TYPE_VOID));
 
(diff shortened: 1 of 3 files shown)


ext/fiddle/fiddle.c
```diff
@@ -164,191 +164,193 @@ Init_fiddle(void)
...
-    /* Document-const: TYPE_INTPTR_T
+    /* Document-const: Fiddle::Types::INTPTR_T
      *
      * C type - intptr_t
      */
-    rb_define_const(mFiddle, "TYPE_INTPTR_T",  INT2NUM(TYPE_INTPTR_T));
+    rb_define_const(mFiddleTypes, "INTPTR_T",  INT2NUM(TYPE_INTPTR_T));
 
-    /* Document-const: TYPE_UINTPTR_T
+    /* Document-const: Fiddle::TypesUINTPTR_T
```
kou, line 349:
```suggestion
    /* Document-const: Fiddle::Types::UINTPTR_T
```

kou (approved):
+1

PR #115, opened 2022-11-22
Add Qtrue, Qfalse, Qnil, and Qundef as constants

tenderlove:
It's easy to get access to `Qtrue`, `Qfalse`, and `Qnil` via Fiddle. You can just do `Fiddle.dlwrap(true)`.  But it's not possible to get access to `Qundef` from pure Ruby.

This patch adds all of these magic constants as constants on Fiddle so that we can write JIT compilers in pure Ruby.


jhawthorn (approved):

k0kubun (approved):
❤️

PR #119, opened 2022-12-27
Update documentation

tenderlove:
The documentation for `Fiddle.dlwrap` and `Fiddle.dlunwrap` were not very accurate and pretty confusing.  This commit updates the documentation so it's easier to understand what the methods do.

### DIFF
--- a/ext/fiddle/fiddle.c
+++ b/ext/fiddle/fiddle.c
@@ -58,18 +58,16 @@ rb_fiddle_free(VALUE self, VALUE addr)
 /*
  * call-seq: Fiddle.dlunwrap(addr)
  *
- * Returns the hexadecimal representation of a memory pointer address +addr+
+ * Returns the Ruby object stored at the memory address +addr+
  *
  * Example:
  *
- *   lib = Fiddle.dlopen('/lib64/libc-2.15.so')
- *   => #<Fiddle::Handle:0x00000001342460>
- *
- *   lib['strcpy'].to_s(16)
- *   => "7f59de6dd240"
- *
- *   Fiddle.dlunwrap(Fiddle.dlwrap(lib['strcpy'].to_s(16)))
- *   => "7f59de6dd240"
+ *    >> x = Object.new
+ *    => #<Object:0x0000000107c7d870>
+ *    >> Fiddle.dlwrap(x)
+ *    => 4425504880
+ *    >> Fiddle.dlunwrap(_)
+ *    => #<Object:0x0000000107c7d870>
  */
 VALUE
 rb_fiddle_ptr2value(VALUE self, VALUE addr)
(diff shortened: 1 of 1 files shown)


ext/fiddle/fiddle.c
```diff
@@ -58,18 +58,16 @@ rb_fiddle_free(VALUE self, VALUE addr)
...
- *   => "7f59de6dd240"
- *
- *   Fiddle.dlunwrap(Fiddle.dlwrap(lib['strcpy'].to_s(16)))
- *   => "7f59de6dd240"
+ *    >> x = Object.new
+ *    => #<Object:0x0000000107c7d870>
+ *    >> Fiddle.dlwrap(x)
+ *    => 4425504880
+ *    >> Fiddle.dlunwrap(_)
+ *    => #<Object:0x0000000107c7d870>
```
zverok, line 70: I believe we prefer examples syntax that is correct Ruby to be copy-pasted in some file and tried (also RDoc's syntax highlight wouldn't work otherwise, so I'd suggest...
```suggestion
 *    x = Object.new
 *    # => #<Object:0x0000000107c7d870>
 *    addr = Fiddle.dlwrap(x)
 *    # => 4425504880
 *    Fiddle.dlunwrap(addr)
 *    # => #<Object:0x0000000107c7d870>
```

zverok, line 70: I believe we prefer examples syntax that is correct Ruby to be copy-pasted in some file and tried (also RDoc's syntax highlight wouldn't work otherwise), so I'd suggest...
```suggestion
 *    x = Object.new
 *    # => #<Object:0x0000000107c7d870>
 *    addr = Fiddle.dlwrap(x)
 *    # => 4425504880
 *    Fiddle.dlunwrap(addr)
 *    # => #<Object:0x0000000107c7d870>
```

ext/fiddle/fiddle.c
```diff
@@ -80,15 +78,22 @@ rb_fiddle_ptr2value(VALUE self, VALUE addr)
...
+ * Returns the memory address of the Ruby object stored at +val+
  *
  * Example:
  *
- *   lib = Fiddle.dlopen('/lib64/libc-2.15.so')
- *   => #<Fiddle::Handle:0x00000001342460>
+ *    >> x = Object.new
+ *    => #<Object:0x0000000107c7d870>
+ *    >> Fiddle.dlwrap(x)
+ *    => 4425504880
```
zverok, line 88:
```suggestion
 *    x = Object.new
 *    # => #<Object:0x0000000107c7d870>
 *    Fiddle.dlwrap(x)
 *    # => 4425504880
```

ext/fiddle/fiddle.c
```diff
@@ -80,15 +78,22 @@ rb_fiddle_ptr2value(VALUE self, VALUE addr)
...
+ *
+ * In the case +val+ is not a heap allocated object, this method will return
+ * the tagged pointer value.
+ *
+ * Example:
  *
- *   Fiddle.dlwrap(lib['strcpy'].to_s(16))
- *   => 25522520
+ *    >> Fiddle.dlwrap(123)
+ *    => 247
```
zverok, line 96:
```suggestion
 *    Fiddle.dlwrap(123)
 *    # => 247
```

ext/fiddle/fiddle.c
```diff
@@ -80,15 +78,22 @@ rb_fiddle_ptr2value(VALUE self, VALUE addr)
...
  *
- *   lib = Fiddle.dlopen('/lib64/libc-2.15.so')
- *   => #<Fiddle::Handle:0x00000001342460>
+ *    >> x = Object.new
+ *    => #<Object:0x0000000107c7d870>
+ *    >> Fiddle.dlwrap(x)
+ *    => 4425504880
+ *
+ * In the case +val+ is not a heap allocated object, this method will return
+ * the tagged pointer value.
```
zverok, line 91: I can't say that "tagged pointer value" says anything to me in the context (and the example doesn't make it much clearer), but that's might be because of my ignorance :)

tenderlove:
@zverok oops, thanks.  I'll update the comments with your suggestions.

I think explaining tagged pointers is probably outside the scope of Fiddle documentation.  I expect developers that are messing around with pointers to Ruby objects need to know a bit about Ruby internals, but I wanted to at least call out that not _all_ Ruby objects passed to `dlwrap` will return a memory address.

PR #134, opened 2023-09-10
Suppress -Wundef warnings on arm64 macOS and Windows

nobu:
```
In file included from ../../../../ext/fiddle/fiddle.h:46:
/opt/local/include/ffi.h:477:5: warning: 'FFI_GO_CLOSURES' is not defined, evaluates to 0 [-Wundef]
    ^
```

c.f. https://github.com/libffi/libffi/pull/796


kou (approved):
+1

PR #136, opened 2024-01-10
Use ruby/actions workflow for ruby versions

m-nakamura145:
this PR has been changed to use the [ruby/actions](https://github.com/ruby/actions/blob/master/.github/workflows/ruby_versions.yml) workflow for ruby versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,15 @@ on:
   - pull_request
 
 jobs:
+  ruby-versions:
+    uses: ruby/actions/.github/workflows/ruby_versions.yml@master
+    with:
+      engine: cruby
+      min_version: 2.5
+      versions: "[debug]"
+
   host:
+    needs: ruby-versions
     name: >-
       ${{ matrix.os }} ${{ matrix.ruby }}
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -5,7 +5,15 @@ on:
   - pull_request
 
 jobs:
+  ruby-versions:
+    uses: ruby/actions/.github/workflows/ruby_versions.yml@master
+    with:
+      engine: cruby
+      min_version: 2.5
+      versions: "[debug]"
```
olleolleolle, line 13: Perhaps like this:

```suggestion
      versions:
        - debug
```

m-nakamura145: Thank you!

kou: It seems that we can't use JSON array here...: https://github.com/ruby/fiddle/actions/runs/7472168103/workflow

```text
The workflow is not valid. .github/workflows/ci.yml (Line: 14, Col: 9): A sequence was not expected
```

m-nakamura145: thanks. fixed [6e2f4b9](https://github.com/ruby/fiddle/pull/136/commits/6e2f4b9957c2aa167bf8924bb33373a71a200fdf)

kou:
We may need to add a condition for `head` to https://github.com/ruby/fiddle/blob/fd4bb4ea8a921985ad1c6a98e787f2dd293d146a/.github/workflows/ci.yml#L51

BTW, do we need both of `head` and `debug`?

PR #138, opened 2024-01-29
Set changelog_uri gem metadata

m-nakamura145:
This PR added [changelog_uri](https://guides.rubygems.org/specification-reference/#metadata) for [rubygems.org](https://rubygems.org/gems/fiddle)


kou:
Thanks.

PR #139, opened 2024-05-04
Ractor support

nobu:


test/fiddle/test_closure.rb
```diff
@@ -149,5 +149,14 @@ def test_memsize_ruby_dev_42480
         end
       end
     end
+
+    if defined?(Ractor)
```
kou, line 153: Could you use `ractor` instead like https://github.com/ruby/pathname/pull/38 ?
We need to run tests that use Ractor after tests that don't use Ractor. It's for running tests that don't use Ractor in single-ractor mode not multi-ractor mode.

nobu: These tests just use `make_shareable` and `shareable?`, and don’t put it in multi-ractor mode, I think.

kou: Ah, you're right.
We don't need to use `ractor`.

How about adding `omit("Need Ractor support") unless defined?(Ractor) to `assert_ractor_shareable`?

ext/fiddle/fiddle.h
```diff
@@ -236,5 +236,15 @@ VALUE rb_fiddle_new_function(VALUE address, VALUE arg_types, VALUE ret_type);
 typedef void (*rb_fiddle_freefunc_t)(void*);
 VALUE rb_fiddle_ptr_new_wrap(void *ptr, long size, rb_fiddle_freefunc_t func, VALUE wrap0, VALUE wrap1);
 
+enum {
+    FIDDLE_DEFAULT_TYPED_DATA_FALGS = (
```
Maumagnaguagno, line 240: s/FALGS/FLAGS

kou: Good catch.
I've fixed it.

PR #140, opened 2024-06-11
Remove Vim commands

tenderlove:
Some of these commands just don't work (for example `sws` is not a Vim setting). I think we should just remove these.

### DIFF
--- a/ext/fiddle/closure.c
+++ b/ext/fiddle/closure.c
@@ -455,4 +455,3 @@ Init_fiddle_closure(void)
      */
     rb_define_method(cFiddleClosure, "freed?", closure_freed_p, 0);
 }
-/* vim: set noet sw=4 sts=4 */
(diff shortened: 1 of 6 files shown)


svardew:
I guess correct modline is ```/* vim: set noet sts=4 st=4: */```

closuer.c, conversions.c:
missing second colon is not valid modeline syntax, vim disable modeline.

fiddle.c, fiddle.h, function.c, handle.c:
sws (swapsync) may be typo of sts (softtabstop).

also, same typo exist in files at ruby/ruby/ext/psych.
```sh
./ext/psych/psych_parser.c:565:/* vim: set noet sws=4 sw=4: */
./ext/psych/psych_yaml_tree.c:12:/* vim: set noet sws=4 sw=4: */
./ext/psych/psych.c:37:/* vim: set noet sws=4 sw=4: */
./ext/psych/psych_emitter.c:555:/* vim: set noet sws=4 sw=4: */
./ext/psych/psych_to_ruby.c:39:/* vim: set noet sws=4 sw=4: */
```

PR #143, opened 2024-08-06
Removed libffi patchs for old Ruby

hsbt:
Pick https://github.com/ruby/ruby/commit/92865d8760e22dc5035a67e636fab3fbd7a77a26 from `ruby/ruby` repo.


kou (approved):
+1

The Fedora CI failure is unrelated to this.
We can fix it by using `docker compose` not `docker-compose` in a separated PR.

PR #144, opened 2024-08-06
Use docker compose instead of docker-compose

hsbt:
Fixed for https://github.com/ruby/fiddle/actions/runs/10260746905/job/28387307591?pr=143


kou:
Can we use `docker compose` instead of `docker-compose`?

hsbt:
👍 I update with that.

Issue #104, opened 2022-01-27
JRuby support


headius:
JRuby ships an implementation of fiddle based on FFI, and cannot support the C extension contained in this gem. It may be possible for us to do what @chrisseaton did and add a native fiddle implementation, but for now we would like to incorporate our own fiddle implementation into this gem, so that we can use it and users can reference it.

kou:
+1

What should we do for it as the next step?

rubyFeedback:
This would be interesting e. g. if we could then use fiddle + jruby for the libui bindings maintained by kojix2. I have not tried this yet (I usually use MRI as-is) but that would be really interesting, even more so with graalvm (if we can manage to compile everything into one app on windows one day; easy-mode distribution for elderly people who may need that. That is one use case I have for java swing, but of course I'd love to use jruby or any ruby-centric application for that).

headius:
I will review the additional changes and we will probably pull this version into JRuby, just not as a gem until JRuby 10.

PR #146, opened 2024-09-17
Skip native extension building on non-CRuby

### LINKED ISSUE #145: The fiddle gem is broken on truffleruby and jruby
Given https://bugs.ruby-lang.org/issues/20309 fiddle will become a bundled gem in Ruby 3.5+.

As a result people will add fiddle to their Gemfile where they previously wouldn't (and so there was no issue).
Unfortunately this leads to failure for example on https://github.com/ffi/ffi/pull/1119:
* https://github.com/ffi/ffi/actions/runs/10881406447/job/30190278321?pr=1119 JRuby fails when trying to build the C extension
* TruffleRuby segfaults, probably because the fiddle C extension does something unexpected, and it's completely untested on TruffleRuby. It seems better to use the TruffleRuby Fiddle backend anyway for performance, as it avoids basically going twice through libffi.

I think the best solution for now is to just use `fiddle` from stdlib on non-CRuby. That's properly tested and working.
IOW, to not build or use the C extension on non-CRuby, similar to how it was done in https://github.com/ruby/stringio/pull/71 or https://github.com/ruby/strscan/pull/35

andrykonchin:
Given https://bugs.ruby-lang.org/issues/20309 fiddle will become a bundled gem in Ruby 3.5+.

As a result people will add fiddle to their Gemfile where they previously wouldn't (and so there was no issue).
Unfortunately this leads to failure for example on https://github.com/ffi/ffi/pull/1119:
* https://github.com/ffi/ffi/actions/runs/10881406447/job/30190278321?pr=1119 JRuby fails when trying to build the C extension
* TruffleRuby segfaults, probably because the fiddle C extension does something unexpected, and it's completely untested on TruffleRuby. It seems better to use the TruffleRuby Fiddle backend anyway for performance, as it avoids basically going twice through libffi (one used for the C extension API and the one used by Fiddle).

The fiddle gem (in this repo) currently only works on CRuby.
So the best solution for now is to just use `fiddle` from stdlib on non-CRuby. That's properly tested and working.

Later it might be possible to upstream the Fiddle backends of JRuby / TruffleRuby in this repo.
But that is a much larger undertaking and we need to fix the failure when people have `gem "fiddle"` in their Gemfile or through a gem's dependencies.

Close #145

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,31 @@ jobs:
           cd tmp
           ruby test/run.rb
 
+  # Don't run tests on TruffleRuby and JRuby because tests don't pass on them yet.
+  # So check only gem installing.
+  install:
+    name: >-
+      Install ${{ matrix.ruby }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby:
+          - truffleruby
+          - jruby
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
+
+      - run: rake install
+
+      - run: ruby -e 'require "fiddle"'
+
   docker:
     name: >-
       ${{ matrix.service }}
--- a/Rakefile
+++ b/Rakefile
@@ -17,8 +17,12 @@ namespace :version do
   end
 end
 
-require 'rake/extensiontask'
-Rake::ExtensionTask.new("fiddle")
-Rake::ExtensionTask.new("-test-/memory_view")
+if RUBY_ENGINE == 'ruby'
+  require 'rake/extensiontask'
+  Rake::ExtensionTask.new("fiddle")
+  Rake::ExtensionTask.new("-test-/memory_view")
+else
+  task :compile
+end
 
 task :default => [:compile, :test]
--- a/lib/fiddle.rb
+++ b/lib/fiddle.rb
@@ -1,6 +1,13 @@
 # frozen_string_literal: true
 
-require 'fiddle.so'
+if RUBY_ENGINE == "ruby"
+  require 'fiddle.so'
+else
+  $LOAD_PATH.delete(__dir__)
+  require 'fiddle' # load from stdlib
+  return
+end
+
 require 'fiddle/closure'
 require 'fiddle/function'
 require 'fiddle/version'
--- a/test/run.rb
+++ b/test/run.rb
@@ -2,7 +2,7 @@
 
 $VERBOSE = true
 
-source_dir = "#{__dir__}/.."
+source_dir = File.dirname(__dir__)
 $LOAD_PATH.unshift("#{source_dir}/test")
 $LOAD_PATH.unshift("#{source_dir}/lib")
 
--- a/ext/fiddle/extconf.rb
+++ b/ext/fiddle/extconf.rb
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 require 'mkmf'
 
+if RUBY_ENGINE != 'ruby'
+  File.write('Makefile', dummy_makefile("").join)
+  return
+end
+
 # :stopdoc:
 
 def gcc?


eregon (approved):
Looks good to me.

@andrykonchin ran the test suite locally and that gave:
```
214 tests, 287 assertions, 13 failures, 107 errors, 0 pendings, 8 omissions, 0 notifications
41.7476% passed
```
So it doesn't make sense to try to add TruffleRuby in CI at this stage, or at least not for running the test suite as it would just fail.

eregon:
Let's add `- { os: ubuntu-latest , ruby: jruby }` as well to make sure the gem installs fine on JRuby too.

.github/workflows/ci.yml
```diff
@@ -58,16 +60,23 @@ jobs:
       - run: gem install pkg/*.gem
 
       - run: rake test
+        # Don't run specs on TruffleRuby and JRuby
```
kou, line 63: "spec" isn't a suitable word for us. "test" is better because we're using test-unit.

Could you add a comment why we don't run tests on TruffleRuby and JRuby?

.github/workflows/ci.yml
```diff
@@ -58,16 +60,23 @@ jobs:
...
+          matrix.ruby != 'jruby'
         env:
           RUBYOPT: --disable=gems
 
       - name: Run test against installed gem
         # We can't use Fiddle gem with RubyInstaller because
         # RubyInstaller requires Fiddle before rubygems is loaded:
         # https://github.com/oneclick/rubyinstaller2/blob/master/lib/ruby_installer/build/dll_directory.rb#L1
+        #
+        # Don't run specs on TruffleRuby and JRuby
```
kou, line 75: ditto.

lib/fiddle.rb
```diff
@@ -1,6 +1,13 @@
 # frozen_string_literal: true
 
-require 'fiddle.so'
+if RUBY_ENGINE == "ruby"
+  require 'fiddle.so'
+else
+  $LOAD_PATH.delete(__dir__)
```
kou, line 6: Why do we need this?

.github/workflows/ci.yml
```diff
@@ -28,6 +28,8 @@ jobs:
         include:
           - { os: windows-latest , ruby: mingw }
           - { os: windows-latest , ruby: mswin }
+          - { os: ubuntu-latest , ruby: truffleruby }
+          - { os: ubuntu-latest , ruby: jruby }
```
kou, line 32: Why do you want to reuse this job?
It seems that you just want to test `rake install` for TruffleRuby and JRuby.

kou (commented):
Could you update the PR description to explain why this PR is needed?

@headius Do you want to proceed #104 for JRuby instead of this approach?

lib/fiddle.rb
eregon: Because we need to load fiddle.rb from the stdlib in TruffleRuby/JRuby, which works, unlike the version of Fiddle in this repo which only works on CRuby.

eregon:
> Could you update the PR description to explain why this PR is needed?

Done.

> @headius Do you want to proceed https://github.com/ruby/fiddle/issues/104 for JRuby instead of this approach?

That can always be done later/separately, as the updated description mentions. We need to fix the immediate issue.

.github/workflows/ci.yml
eregon: My idea is to do that to avoid duplicating lines 38 to 60, to ensure they stay in sync.
And it also makes it easier to later enable tests e.g. for TruffleRuby once all or almost all Fiddle tests pass.

Given that, is it OK as-is or you prefer a separate job with some duplication?

.github/workflows/ci.yml
andrykonchin: Done

.github/workflows/ci.yml
andrykonchin: Done

headius:
@kou We do still intend to import our Fiddle implementation into this gem, as will all other stdlib gems. That is not a high priority at the moment, so it's ok to move forward with this PR for now.

.github/workflows/ci.yml
kou: Do we need to duplicate lines 38 to 60?

It seems that we just need this:

```yaml
      - uses: actions/checkout@v4
      - name: Set up Ruby
        uses: ruby/setup-ruby@v1
        with:
          ruby-version: ${{ matrix.ruby }}
          bundler-cache: true
      - run: bundle exec rake install
      - run: ruby -e 'require "fiddle"'
```

> And it also makes it easier to later enable tests e.g. for TruffleRuby once all or almost all Fiddle tests pass.

It there a plan to work on this? If there isn't a plan for this, we don't need to care about this.

> Given that, is it OK as-is or you prefer a separate job with some duplication?

Could you create a separate job? We don't want to maintain `if:` as much as possible.

lib/fiddle.rb
kou: I don't think that changing `$LOAD_PATH` isn't a good approach because `$LOAD_PATH` is one of global variables.
Can we use another approach?

kou (commented):
> We do still intend to import our Fiddle implementation into this gem, as with all other stdlib gems. That is not a high priority at the moment, so it's ok to move forward with this PR for now.

OK. Thanks for sharing your thought.

lib/fiddle.rb
eregon: There is no other way (AFAIK). We need to make sure the Ruby files of this gem are not loaded because it's a different version than in stdlib. So we need to remove it from $LOAD_PATH.
It is the same approach as for older FFI versions: https://github.com/ffi/ffi/blob/85d0fabce6ab5ceb3848f7e09a265341672a272d/lib/ffi.rb#L19

.github/workflows/ci.yml
eregon: Alright, let's use a separate job. I'm unsure if we need the version bump as well.

andrykonchin: Done

lib/fiddle.rb
kou: I feel that it's an ad-hoc approach.
Can we use https://github.com/oracle/truffleruby/blob/512427a658f29b222df9a6da1b3ffc12b46a3be6/lib/mri/fiddle.rb#L3-L4 ?

test/run.rb
```diff
@@ -3,8 +3,8 @@
 $VERBOSE = true
 
 source_dir = "#{__dir__}/.."
-$LOAD_PATH.unshift("#{source_dir}/test")
-$LOAD_PATH.unshift("#{source_dir}/lib")
+$LOAD_PATH.unshift(File.expand_path("#{source_dir}/test"))
+$LOAD_PATH.unshift(File.expand_path("#{source_dir}/lib"))
```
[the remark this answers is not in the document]
andrykonchin: Yes, it's needed to be able to remove `.../fiddle/lib` path from `$LOAD_PATH` in `lib/fiddle.rb`:

```ruby
  $LOAD_PATH.delete(__dir__)
```

The `source_dir` absolute path isn't normalised and contains `"#{__dir__}/.."`.

The first line for the `test` directory is changes for consistency.

.github/workflows/ci.yml
```diff
@@ -73,6 +73,36 @@ jobs:
...
+          - jruby
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+
+      - run: bundle install
```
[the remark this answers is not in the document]
andrykonchin: Done

.github/workflows/ci.yml
```diff
@@ -73,6 +73,36 @@ jobs:
...
+
+      - run: rake compile
+
+      - run: rake build
+
+      - run: gem install pkg/*.gem
```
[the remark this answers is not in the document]
andrykonchin: Done

.github/workflows/ci.yml
```diff
@@ -73,6 +73,36 @@ jobs:
           cd tmp
           ruby test/run.rb
 
+  # Don't run tests on TruffleRuby and JRuby because tests don't pass on them yet.
+  # So check only compiling.
+  host-with-non-CRuby:
```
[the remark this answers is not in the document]
andrykonchin: Done

.github/workflows/ci.yml
```diff
@@ -73,6 +73,36 @@ jobs:
...
+    name: >-
+      ${{ matrix.os }} ${{ matrix.ruby }}
```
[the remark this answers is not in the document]
andrykonchin: Done

.github/workflows/ci.yml
```diff
@@ -73,6 +73,36 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
```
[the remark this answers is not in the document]
andrykonchin: Done

.github/workflows/ci.yml
```diff
@@ -73,6 +73,36 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
```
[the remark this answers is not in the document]
andrykonchin: Done

test/run.rb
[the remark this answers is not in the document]
kou: I think that this file isn't used by TruffleRuby.
This file is used to run tests but TruffleRuby doesn't run tests.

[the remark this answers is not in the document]
eregon: Well, that change is useful to run Fiddle tests locally on TruffleRuby, is there any problem with it?

lib/fiddle.rb
eregon: > I feel that it's an ad-hoc approach.

It is the only approach that reliably works (AFAIK).

> Can we use https://github.com/oracle/truffleruby/blob/512427a658f29b222df9a6da1b3ffc12b46a3be6/lib/mri/fiddle.rb#L3-L4 ?

No, that wouldn't work because then we would mix this gem's version of the Ruby files with the backend files expecting a different Fiddle version.
As an example if any method moved between the Ruby files and the C extension here it would break. Or if the Ruby files expect some method to be defined in the C extension that would also break.
In addition since TruffleRuby is not yet in this repo's CI it would be very likely to break silently.

When TruffleRuby passes all/almost-all Fiddle tests we will consider upstreaming what makes sense and analyze what's the best interface (I'm thinking either upstream the pure-Ruby backend or use the C extension's methods as the interface but then it needs to be more stable and won't be very practical. or define a separate interface between Fiddle frontend & backend). It is too early for that since many tests don't pass.

kou: > > I feel that it's an ad-hoc approach.
> 
> It is the only approach that reliably works (AFAIK).

This approach depends on `$LOAD_PATH` contents. So you want to add `test/run.rb` changes, right?

Can we prepare an empty fiddle gem only for TruffleRuby?

test/run.rb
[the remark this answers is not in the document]
kou: Not adding `lib/` to `$LOAD_PATH` is better for the purpose.

Anyway, changing `test/run.rb` should be out-of-scope of this PR. We should work on it as a separated work.

lib/fiddle.rb
eregon: > This approach depends on `$LOAD_PATH` contents.

Yes, there is no other way for `require 'fiddle'` anywhere in the program to load from stdlib and not from the gem (AFAIK).

> So you want to add `test/run.rb` changes, right?

That file is adding "absolute" non-normalized paths to `$LOAD_PATH`, so it's just fixing that. Applications typically don't have this issue as the `$LOAD_PATH` entry for the fiddle gem will be added by RubyGems or Bundler.

> Can we prepare an empty fiddle gem only for TruffleRuby?

There is no TruffleRuby-specific `Gem::Platform` (https://github.com/rubygems/rubygems/blob/master/lib/rubygems/platform.rb).
Also this issue of fiddle becoming a bundled gem affects multiple Ruby implementations, so it's not just about TruffleRuby but also JRuby and maybe others.
(and it's also not just about fiddle, it's the same for all default->bundled gems transitions when the gem has a C extension and the extension doesn't just work on these Rubies).
Reusing the stdlib is the obvious short-term fix because that's what was used before and known to be working & tested in that Ruby implementation.
In short, it unbreaks what the warnings in https://bugs.ruby-lang.org/issues/20309 broke.
Longer-term these Ruby implementations might be able to upstream their Fiddle backend in this repo.

test/run.rb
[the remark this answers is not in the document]
eregon: Alright, let's undo that change since it's not strictly necessary in this PR.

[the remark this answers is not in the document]
andrykonchin: Done

.github/workflows/ci.yml
```diff
@@ -73,6 +73,29 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
+
+      - run: rake install
```
eregon, line 97:
```suggestion
      - run: rake install
      - run: ruby -e 'require "fiddle"'
```
like @kou suggested in https://github.com/ruby/fiddle/pull/146#discussion_r1765898874 seems good to at least check it requires successfully.

lib/fiddle.rb
kou: Can we prepend the system path to `$LOAD_PATH` instead of deleting a path from `$LOAD_PATH` instead something like the following?

```ruby
system_fiddle_lib_dir = RbConfig::CONFIG["rubylibdir"]
begin
  load_path = $LOAD_PATH.dup
  $LOAD_PATH.prepend(system_fiddle_lib_dir)
  require "fiddle"
ensure
  # The above 'require "fiddle"' must not change $LOAD_PATH but...
  $LOAD_PATH.shift if $LOAD_PATH[0] == system_fiddle_lib_dir
end
```

kou: Can we use `require "#{system_fiddle_lib_dir}/fiddle"` and `require_relative` in TruffleRuby's Fiddle?

eregon: Both of these are far more complicated (and are likely broken for some cases, e.g. `require "fiddle"` might repetitively change $LOAD_PATH for the former) than the change in this PR.
What's the concrete problem with that change? It's simple and it does the job.

eregon: > Can we use `require "#{system_fiddle_lib_dir}/fiddle"` and `require_relative` in TruffleRuby's Fiddle?

It wouldn't work if user code does e.g. `require 'fiddle/closure'`.
And changing fiddle files in TruffleRuby is increasing the diff + it does not fix the issue for existing TruffleRuby releases.

kou: > What's the concrete problem with that change?

I don't want to change a global variable as I already said.

> It wouldn't work if user code does e.g. `require 'fiddle/closure'`.

`require 'fiddle/closure'` doesn't work with your `$LOAD_PATH.delete` approach.

BTW, why do users need to use `require "fiddle/closure"`? `fiddle.rb` requires `fiddle/closure.rb` automatically.

> changing fiddle files in TruffleRuby is increasing the diff to upstream

ruby/fiddle also uses `require_relative`.

> it does not fix the issue for existing TruffleRuby releases.

Can users pin old gems that don't have explicit `fiddle` dependencies for the case?

eregon: > I don't want to change a global variable as I already said.

It's not Fiddle's global variable.
`$LOAD_PATH` is the only way to change what `require "fiddle"` finds (AFAIK), so it's like part of the "API" for `require`.
We are basically undoing what RubyGems did by adding the Fiddle gem to `$LOAD_PATH`, which is exactly what we want here as the only thing that works is using the stdlib.

> `require 'fiddle/closure'` doesn't work with your `$LOAD_PATH.delete` approach.

It does if the order is `require 'fiddle'` and then `require 'fiddle/closure'` (in user code).
But indeed it doesn't if it's `require 'fiddle/closure'` (in user code) before `require 'fiddle'`.

> BTW, why do users need to use `require "fiddle/closure"`? `fiddle.rb` requires `fiddle/closure.rb` automatically.

Agreed, they shouldn't. It illustrates that changing `$LOAD_PATH` just for the duration of that require is brittle.

> ruby/fiddle also uses `require_relative`.

No it doesn't:
https://github.com/ruby/fiddle/blob/aad5a3bc797c7271bbc06258fe5cd99cd3f3d2bb/lib/fiddle.rb#L4-L6

> > it does not fix the issue for existing TruffleRuby releases.
> 
> Can users pin old gems that don't have explicit `fiddle` dependencies for the case?

That seems too complicated and impractical. Many things in the Ruby ecosystem automatically update to the latest gem versions, etc. Also the idea is an application running on CRuby should run on TruffleRuby as-is, without changing the Gemfile (because doing so might then bring back some bugs in gems, etc). Having to figure out this kind of version constraints is a nightmare and too impractical.

eregon: @hsbt In https://bugs.ruby-lang.org/issues/18567#note-35 you said:
> I can help default/bundled gems are working with JRuby/TruffleRuby like:

I would like to ask your help here in convincing @kou this is the way to go, or to find an alternative if you know one.
I don't see any other solution which would work on existing TruffleRuby & JRuby releases.
https://bugs.ruby-lang.org/issues/20309 is breaking TruffleRuby & JRuby, this PR is AFAIK the only short-term solution.
This is BTW nothing new, it's the approach I mentioned in https://bugs.ruby-lang.org/issues/18567#note-30 (for strscan we eventually found an approach of not shipping a `strscan.rb` in the gem, but that's not an option here).

kou: > It's not Fiddle's global variable.

There is no "XXX's global variable". Because it's "global".
I said that I don't want to change any global variables by `require "fiddle"`.

> `$LOAD_PATH` is the only way to change what `require "fiddle"` finds (AFAIK)

`require "#{system_fiddle_lib_dir}/fiddle" and `require_relative` can change it as I already mentioned.

> > BTW, why do users need to use `require "fiddle/closure"`? `fiddle.rb` requires `fiddle/closure.rb` automatically.
> 
> Agreed, they shouldn't. It illustrates that changing `$LOAD_PATH` just for the duration of that require is brittle.

We don't need to care about unexpected cases.
It shows that the approach doesn't have real world problem.

> > ruby/fiddle also uses `require_relative`.
> 
> No it doesn't:

Ah, sorry. I wanted to say that we can change the current `require` to `require_relative` in ruby/fiddle. So "increasing the diff to upstream" is not a blocker.

> > > it does not fix the issue for existing TruffleRuby releases.
> > 
> > 
> > Can users pin old gems that don't have explicit `fiddle` dependencies for the case?
> 
> That seems too complicated and impractical. Many things in the Ruby ecosystem automatically update to the latest gem versions, etc. Also the idea is an application running on CRuby should run on TruffleRuby as-is, without changing the Gemfile (because doing so might then bring back some bugs in gems, etc). Having to figure out this kind of version constraints is a nightmare and too impractical.

If TruffleRuby works on the real fix ("Longer-term these Ruby implementations might be able to upstream their Fiddle backend in this repo") as soon as possible, we can avoid this situation.

If this is a really short-term fix, we don't need `require_relative` changes. Because `lib/fiddle/**/*.rb` in ruby/fiddle and TruffleRuby will not be different. So existing TruffleRuby releases will work with `lib/fiddle/**/*.rb` in ruby/fiddle.

Rakefile
```diff
@@ -17,8 +17,12 @@ namespace :version do
...
-require 'rake/extensiontask'
-Rake::ExtensionTask.new("fiddle")
-Rake::ExtensionTask.new("-test-/memory_view")
+if RUBY_ENGINE == 'ruby'
+  require 'rake/extensiontask'
+  Rake::ExtensionTask.new("fiddle")
+  Rake::ExtensionTask.new("-test-/memory_view")
+else
+  task :compile
+end
```
kou, line 26: Can we revert this?
We don't need to use `rake default` with TruffleRuby.

lib/fiddle.rb
eregon: > I said that I don't want to change any global variables by require "fiddle".

That wasn't clear to me, and I guess you mean change any global variable permanently, and that's why you prefer https://github.com/ruby/fiddle/pull/146#discussion_r1772483803 to just `$LOAD_PATH.delete(__dir__)`, correct?

I think we can try that solution, I will need to check it behaves correctly for multiple `require 'fiddle'` calls.

> I wanted to say that we can change the current require to require_relative in ruby/fiddle. So "increasing the diff to upstream" is not a blocker.

Right, it's fine to change it here, but it doesn't solve the issue for existing truffleruby & jruby releases.

> If TruffleRuby works on the real fix ("Longer-term these Ruby implementations might be able to upstream their Fiddle backend in this repo") as soon as possible, we can avoid this situation.

There are many such gems which became bundled gems, so it's not an issue only for fiddle but for [a number of gems](https://github.com/oracle/truffleruby/issues/2644). For some of these gems (but not all) we can just reuse the C extension on TruffleRuby though, so for those gems it's then only a problem for JRuby.

In terms of priority for upstreaming TruffleRuby backend into the respective gems, `strscan` feels higher than `fiddle` because there seem to be more frequent changes there.
Anyway, we will get to it, but to solve it for existing releases we need this PR.

> If this is a really short-term fix, we don't need require_relative changes. Because `lib/fiddle/**/*.rb` in ruby/fiddle and TruffleRuby will not be different. So existing TruffleRuby releases will work with `lib/fiddle/**/*.rb` in ruby/fiddle.

This is too brittle and as I already said it would break silently.
The next TruffleRuby release is in 6 months, I don't think it's a good idea to require no changes in Fiddle Ruby files for 6 months.
Also it doesn't solve it for JRuby.

---

@kou Let's try https://github.com/ruby/fiddle/pull/146#discussion_r1772483803, are you OK with that for now?

Rakefile
eregon: On one side you seem to be asking a clean integration by upstraming the Fiddle TruffleRuby here soon, and on the other side you seem to dislike any change making incremental progress towards that.
I don't see any point to revert this, it's useful to run fiddle tests on TruffleRuby.

lib/fiddle.rb
eregon: I tried https://github.com/ruby/fiddle/pull/146#discussion_r1772483803 after `rake install` and with `ruby -ve 'require "fiddle"; puts $"; puts; require "fiddle"; puts $"'`, it seems to work fine.

I pushed a commit switching to that approach, I hope that addresses your concerns.

eregon: Actually that approach is broken, as can be seen when trying to run the test suite, because `fiddle.rb` does not load all fiddle Ruby files but only some of them. For example `fiddle/pack.rb` isn't loaded eagerly.
And so when the user or tests does `require 'fiddle/struct'` we'll end up with a mix of files from stdlib and from the gem:
```
/home/eregon/.rubies/truffleruby-24.1.0/lib/truffle/truffle/fiddle_backend.rb
/home/eregon/.rubies/truffleruby-24.1.0/lib/mri/fiddle/closure.rb
/home/eregon/.rubies/truffleruby-24.1.0/lib/mri/fiddle/function.rb
/home/eregon/.rubies/truffleruby-24.1.0/lib/mri/fiddle/version.rb
/home/eregon/.rubies/truffleruby-24.1.0/lib/mri/fiddle.rb
/home/eregon/code/fiddle/lib/fiddle.rb
/home/eregon/code/fiddle/test/fiddle/helper.rb
/home/eregon/code/fiddle/lib/fiddle/value.rb
```

So I will remove the last commit.
`$LOAD_PATH.delete(__dir__)` is the only solution so far which works.

eregon: > If this is a really short-term fix, we don't need `require_relative` changes. Because `lib/fiddle/**/*.rb` in ruby/fiddle and TruffleRuby will not be different. So existing TruffleRuby releases will work with `lib/fiddle/**/*.rb` in ruby/fiddle.

For completeness I tried and it doesn't work, the test suite fails during loading and just `ruby -Ilib -e 'require "fiddle"; require "fiddle/struct"'` fails with:
```
$ ruby -Ilib -e 'require "fiddle"; require "fiddle/struct"'
/home/eregon/code/fiddle/lib/fiddle/pack.rb:18:in `const_missing': uninitialized constant Fiddle::PackInfo::TYPE_BOOL (NameError)
Did you mean?  Fiddle::TYPE_LONG
               Fiddle::TYPE_VOID
	from /home/eregon/code/fiddle/lib/fiddle/pack.rb:18:in `<module:PackInfo>'
	from /home/eregon/code/fiddle/lib/fiddle/pack.rb:5:in `<module:Fiddle>'
	from /home/eregon/code/fiddle/lib/fiddle/pack.rb:4:in `<top (required)>'
	from <internal:core> core/kernel.rb:255:in `require'
	from /home/eregon/code/fiddle/lib/fiddle/struct.rb:4:in `<top (required)>'
	from <internal:core> core/kernel.rb:255:in `require'
	from -e:1:in `<main>'
```
OTOH `$LOAD_PATH.delete(__dir__)` works for that `ruby` command and the test suite loads and runs all tests (not all passing, but at least they all load).

kou: > > I said that I don't want to change any global variables by require "fiddle".
> 
> That wasn't clear to me, and I guess you mean change any global variable permanently, and that's why you prefer [#146 (comment)](https://github.com/ruby/fiddle/pull/146#discussion_r1772483803) to just `$LOAD_PATH.delete(__dir__)`, correct?

Ah, I understand. Correct.
 
> This is too brittle and as I already said it would break silently.

We can detect it by adding more checks to CI. For example, `ruby -e 'require "fiddle"; require "fiddle/struct"'` instead of `ruby -e 'require "fiddle"; require "fiddle/struct"'` as you showed.

> The next TruffleRuby release is in 6 months, I don't think it's a good idea to require no changes in Fiddle Ruby files for 6 months.

We can use TruffleRuby compatible change something like if we have a CI job for TruffleRuby:

```diff
diff --git a/lib/fiddle/pack.rb b/lib/fiddle/pack.rb
index 81088f4..99e0742 100644
--- a/lib/fiddle/pack.rb
+++ b/lib/fiddle/pack.rb
@@ -15,8 +15,11 @@ module Fiddle
       TYPE_USHORT => ALIGN_SHORT,
       TYPE_UINT   => ALIGN_INT,
       TYPE_ULONG  => ALIGN_LONG,
-      TYPE_BOOL   => ALIGN_BOOL,
     }
+    # TruffleRuby doesn't have Fiddle::TYPE_BOOL yet.
+    if Fiddle.const_defined?(:TYPE_BOOL)
+      ALIGN_MAP[TYPE_BOOL] = ALIGN_BOOL
+    end
 
     PACK_MAP = {
       TYPE_VOIDP => "L!",
@@ -31,15 +34,18 @@ module Fiddle
       TYPE_UINT   => "I!",
       TYPE_ULONG  => "L!",
     }
-    case SIZEOF_BOOL
-    when SIZEOF_CHAR
-      PACK_MAP[TYPE_BOOL] = PACK_MAP[TYPE_UCHAR]
-    when SIZEOF_SHORT
-      PACK_MAP[TYPE_BOOL] = PACK_MAP[TYPE_USHORT]
-    when SIZEOF_INT
-      PACK_MAP[TYPE_BOOL] = PACK_MAP[TYPE_UINT]
-    when SIZEOF_LONG
-      PACK_MAP[TYPE_BOOL] = PACK_MAP[TYPE_ULONG]
+    # TruffleRuby doesn't have Fiddle::SIZEOF_BOOL yet.
+    if Fiddle.const_defined?(:SIZEOF_BOOL)
+      case SIZEOF_BOOL
+      when SIZEOF_CHAR
+        PACK_MAP[TYPE_BOOL] = PACK_MAP[TYPE_UCHAR]
+      when SIZEOF_SHORT
+        PACK_MAP[TYPE_BOOL] = PACK_MAP[TYPE_USHORT]
+      when SIZEOF_INT
+        PACK_MAP[TYPE_BOOL] = PACK_MAP[TYPE_UINT]
+      when SIZEOF_LONG
+        PACK_MAP[TYPE_BOOL] = PACK_MAP[TYPE_ULONG]
+      end
     end
 
     SIZE_MAP = {
@@ -54,8 +60,11 @@ module Fiddle
       TYPE_USHORT => SIZEOF_SHORT,
       TYPE_UINT   => SIZEOF_INT,
       TYPE_ULONG  => SIZEOF_LONG,
-      TYPE_BOOL   => SIZEOF_BOOL,
     }
+    # TruffleRuby doesn't have Fiddle::TYPE_BOOL yet.
+    if Fiddle.const_defined?(:TYPE_BOOL)
+      SIZE_MAP[TYPE_BOOL] = SIZEOF_BOOL
+    end
     if defined?(TYPE_LONG_LONG)
       ALIGN_MAP[TYPE_LONG_LONG] = ALIGN_MAP[TYPE_ULONG_LONG] = ALIGN_LONG_LONG
       PACK_MAP[TYPE_LONG_LONG] = "q"
```

> Also it doesn't solve it for JRuby.

We can release an empty gem for JRuby like other gems did.

> Actually that approach is broken, as can be seen when trying to run the test suite, because `fiddle.rb` does not load all fiddle Ruby files but only some of them. For example `fiddle/pack.rb` isn't loaded eagerly. And so when the user or tests does `require 'fiddle/struct'` we'll end up with a mix of files from stdlib and from the gem:

How about requiring `fiddle/struct` in `fiddle.rb`?

```ruby
  stdlib = RbConfig::CONFIG["rubylibdir"]
  $LOAD_PATH.prepend(stdlib)
  begin
    require 'fiddle'
    require 'fiddle/struct'
  ensure
    if $LOAD_PATH.first == stdlib
      $LOAD_PATH.shift
    else
      raise 'fiddle files unexpectedly modified the $LOAD_PATH'
    end
  end
```

Or we can update `fiddle/pack.rb` in ruby/fiddle to TruffleRuby compatible as I showed.

Rakefile
kou: I just want to keep 1 logical change in 1 PR. Is it strange?
I think that this PR focus on "workaround for bundled gem" not "upstreaming TruffleRuby implementation".

> the changes in `test/run.rb` is similar, it makes it easier to run tests in order to make them pass

It's needless when we don't use `$LOAD_PATH.delete` approach.

lib/fiddle.rb
eregon: @kou I am not sure why you are trying to suggest even more complex or brittle solutions.
(brittle because `ruby -e 'require "fiddle"; require "fiddle/struct"'` is not proper testing and only covers one example, there are other fiddle files that are not eagerly loaded)
I have shown a solution that works and I have shown the various solutions your proposed so far do not work.
That solution is generic, i.e. it works for other gems too, not just fiddle, and we'll need to do something similar for some other default/bundled extension gems too.
So I have no interest to have highly-fiddle-specific solution here, it's just making things more complicated and less consistent.

Also the solution I propose is effectively equivalent to what is already done in `strscan` and `stringio`, where there it works because there is no `.rb` file, which is conceptually equivalent to not adding the gem `lib/` dir to `$LOAD_PATH`.

It is natural for language implementations to reimplement some of the core & standard library based on what makes sense for them, and possibly upstream that if it is stable enough and does not rely on too many internals. This upstreaming is a significant time investment, it does not happen in days but typically takes weeks or months (as we have already seen). It also significantly hinders the ability of language implementations to optimize these parts of stdlib better because then that upstreamed code needs to care about compatibility e.g. with older TruffleRuby versions, etc.

What is a concrete objective problem with `$LOAD_PATH.delete(__dir__)`?

It seems clear that "$LOAD_PATH is a global variable and shouldn't be changed" is not a good enough argument to reject that approach, `$LOAD_PATH` is part of the require API, it's the only way to influence where `require` looks. It is the natural way to solve this issue (short-term), where the gem is unexpectedly shadowing the standard library (which worked fine until now, as very few gems/Gemfile depended on `fiddle`).

> We can release an empty gem for JRuby like other gems did.

How is this any better? Am I missing something?
As far as I understand, it would require you to do more work by building + pushing + testing the extra java-platform gem (if forgotten it breaks things) and it's effectively exactly the same as `$LOAD_PATH.delete(__dir__)`.

And https://github.com/ruby/fiddle/issues/104 would also be more work for you as then you'd need to maintain a Java implementation.

This PR, once merged & released should require no work or effort from you at all. I thought you would like that.

Also to give you my point of view:
* CRuby broke some stdlib/default/bundled gems for other Ruby implementations by making them bundled gems (or adding warnings for them)
* I filed the issue in https://github.com/ruby/fiddle/issues/145
* @andrykonchin and I created a fix in this PR
* Now we are arguing for days about how it should be done, it is frustrating. It must be for you as well, I'm sorry for that.
* I don't have much time left to spend on this.

If there is still no progress on this PR, I guess one approach is to add some kind of patch in TruffleRuby to do `$LOAD_PATH.delete(fiddle_gem_lib_dir)` when `require 'fiddle'` to never load fiddle from the gem.
I don't think that's a good approach/solution though because:
* It won't fix existing TruffleRuby & JRuby releases.
* It's hidden vs being clear here in the gem.

kou: > I am not sure why you are suggesting more complex or brittle solutions.

I can't understand why I need to say the same thing again and again:
I don't want to change any global variables by require "fiddle".

I can't understand why you say that it's not clear. 

>  (brittle because `ruby -e 'require "fiddle"; require "fiddle/struct"'` is not proper testing and only covers one example, there are other fiddle files that are not eagerly loaded) I have shown a solution that works reliably and I have shown the various solutions your proposed so far do not work.

I already showed a solution for it. We can use it for other files.

> That solution is generic, i.e. it works for other gems too, not just fiddle, and we'll need to do something similar for some other default/bundled extension gems too. So I have no interest to have highly-fiddle-specific solution here, it's just making things more complicated and less consistent (but if it would work reliably, fine, I'd be OK with it).
>
> Also the solution I propose is effectively equivalent to what is already done in `strscan` and `stringio`, where there it works because there is no `.rb` file, which is conceptually equivalent to not adding the gem `lib/` dir to `$LOAD_PATH`.

In principal, your solution (ignore gem) isn't a good solution. If an user wants to use the specific version of A gem, the version information is just ignored with TruffleRuby. It just avoids `require` error. I pointed out this 2 years ago: https://github.com/ruby/strscan/pull/35#discussion_r856753083

It seems that you say that this is a "short-term" workaround but you don't think that this is a "short-term" workaround. At least strscan still uses this workaround. I don't think that it's "short-term".

Anyway, here are PRs that import JRuby and TruffleRuby implementations:

* GH-147
* GH-148

hsbt:
Sorry, I read the detail of this thread today. I'm +1 to @kou 's direction. I prefer #147 and #148.

lib/fiddle.rb
eregon: > I don't want to change any global variables by require "fiddle".

But without that, there is AFAIK no short-term solution to make it work.
Hence while I understand you don't like to change `$LOAD_PATH`, since it is the only solution to fix existing releases (broken indirectly by the fiddle warning added in CRuby), I think it is a reasonable short-term solution.

> I already showed a solution for it. We can use it for other files.

It does feel brittle as if a new file is added it could be missed there. I suppose we could use `Dir.glob` but anyway that snippet of code is just more complicated and feels more risky than `$LOAD_PATH.delete(__dir__)`.
With `$LOAD_PATH.delete(__dir__)` the behavior is clear, with the alternatives it's far more subtle and therefore prone to bugs.

> In principal, your solution (ignore gem) isn't a good solution. If an user wants to use the specific version of A gem, the version information is just ignored with TruffleRuby. 

Yes that's a valid point.
However, a subset of the functionality working is much better than nothing working or even `require 'fiddle'` failing.
The tension here is with:
> It also significantly hinders the ability of language implementations to optimize these parts of stdlib better because then that upstreamed code needs to care about compatibility e.g. with older TruffleRuby versions, etc.

And for example we found that synchronization for `stringio` is a huge overhead, and optimizing that would be much harder if TruffleRuby's stringio implementation lives in ruby/stringio and needs to support multiple truffleruby versions.

> It seems that you say that this is a "short-term" workaround but you don't think that this is a "short-term" workaround. At least strscan still uses this workaround. I don't think that it's "short-term".

Actually on September 12 I wrote to @andrykonchin in a private chat:
> mmh strscan is adding some new features and they are not added for truffleruby since we have `strscan.rb` in the truffleruby repo: https://github.com/ruby/strscan/pull/106
> Probably we should consider moving that upstream. We need to look at the stability of the Primitive's used there though, and Truffle:: methods

And there is the same concern with moving the code here.

> Anyway, here are PRs that import JRuby and TruffleRuby implementations:

Thank you, I didn't expect that.
There are some concerns we need to discuss before that can be merged, I'll reply on that PR.

headius:
@kou Thanks so much for helping to import the JRuby FFI-based version of fiddle! I will comment and help starting today.

kou:
We can use #148 as a short-term solution by requiring only "existing TruffleRuby releases work with `gem "fiddle"` like built-in Fiddle".
If we require many things such as faster implementation, better TruffleRuby compatibility and so on, it may not be a short-term solution. But it's your decision. (I think that we can improve upstreamed TruffleRuby implementation step-by-step after we release a new version with JRuby and TruffleRuby implementations.)

Note that I've implemented many "NotImplemented" methods in TruffleRuby's Fiddle. So #148 has better compatibility than TruffleRuby's Fiddle.

We don't use this as a short-term solution. I don't want to continue the "whether the `$LOAD_PATH.delete` approach is the best or not" discussion. I think that we can't reach consensus for it.

If you don't want to choose #148 as a short-term solution (you want to improve #148 as the best implementation), we just ship only JRuby implementation and release a new version for `gem "fiddle"` cases.

eregon:
#148 is not a short-term solution, we need to review and change every usage of private APIs there.
It's like if Fiddle would use internal CRuby functions not in headers (which might disappear/be renamed in the next version, etc), that would be unacceptable, it's the same here.

We had a similar situation with the FFI gem and what has worked very well is up to some TruffleRuby version use the stdlib (more precisely the version of ffi bundled with TruffleRuby), and for newer versions use the proper integration: https://github.com/ffi/ffi/blob/c128cede750242fe19945af8bd6c797728489ad5/lib/ffi.rb#L10-L27

Also note there the interface is "the methods the C extension exposes, no more, no less".
So there is no concern of using private APIs in the FFI gem, the only "API" used there is `require 'truffleruby/ffi_backend'`.

https://github.com/ruby/fiddle/pull/147 for both JRuby and TruffleRuby could maybe be a shorter-term solution, but it will take some time to relicense.
Also that may be less complete than the Fiddle support in TruffleRuby stdlib (I'm not sure).

> I don't want to continue the "whether the $LOAD_PATH.delete approach is the best or not" discussion. I think that we can't reach consensus for it.

Then maybe we should pick another approach, like the diff in https://github.com/ruby/fiddle/pull/146#discussion_r1777921796 + some testing in CI that basic things work (e.g. require all files, and some basic Fiddle usage to call a libc function).
It is more fragile, but better than nothing working.

> But it's your decision.

I disagree, you are very much forcing my hand here, by making me choose between:
* Any application depending on fiddle (and there will be more since CRuby warns for it) does not work on the latest TruffleRuby release
* Give up completely on what is private or public APIs of TruffleRuby and forever bear the cost of it

So I would like something which is not extreme as that. I think that's a reasonable request.
IOW, I want a solution for https://github.com/ruby/fiddle/issues/145 that works on existing TruffleRuby releases and which does not involve giving up on what is private or public APIs of TruffleRuby.

Would you accept a PR using the diff in https://github.com/ruby/fiddle/pull/146#discussion_r1777921796 and adding basic tests to run on TruffleRuby releases? (we can also try to reuse existing tests but that will likely be a lot of `skip`s)

kou:
> It's like if Fiddle would use internal CRuby functions not in headers (which might disappear/be renamed in the next version, etc), that would be unacceptable, it's the same here.

We'll use `have_func` in `extconf.rb` or something for the case. We may use `respond_to?` or something in `.rb` too.

> #147 for both JRuby and TruffleRuby could maybe be a shorter-term solution, but it will take some time to relicense.

#147 isn't a short-term solution for JRuby. It's a real solution for JRuby.

Relicense has done.

> Also that may be less complete than the Fiddle support in TruffleRuby stdlib (I'm not sure).

#147 is more complete than the Fiddle in TruffleRuby. For example, #147 has variable arguments support but the Fiddle in TruffleRuby doesn't have.
 
> > I don't want to continue the "whether the $LOAD_PATH.delete approach is the best or not" discussion. I think that we can't reach consensus for it.
> 
> Then maybe we should pick another approach, like the diff in [#146 (comment)](https://github.com/ruby/fiddle/pull/146#discussion_r1777921796) + some testing in CI that basic things work (e.g. require all files, and some basic Fiddle usage to call a libc function). It is more fragile, but better than nothing working.

It was an option but it is not an option now because we have #147 and #148. 

> * Give up completely on what is private or public APIs of TruffleRuby and forever bear the cost of it

You can defer it by just adding a comment something like the following:

```diff
diff --git a/lib/fiddle/truffleruby.rb b/lib/fiddle/truffleruby.rb
index f1c34c7..c49a72d 100644
--- a/lib/fiddle/truffleruby.rb
+++ b/lib/fiddle/truffleruby.rb
@@ -9,6 +9,9 @@
 # GNU General Public License version 2, or
 # GNU Lesser General Public License version 2.1.
 
+# CAUTION: This uses private APIs temporary. You should not do this in
+# other files.
+
 module Truffle::FiddleBackend
 
   SIZEOF_CHAR   = Primitive.pointer_find_type_size(:char)
```

If you want to use it as a long-term workaround, we can also use different file entirely for each TruffleRuby release to support multiple TruffleRuby releases something like:

```diff
diff --git a/lib/fiddle/truffleruby.rb b/lib/fiddle/truffleruby.rb
index f1c34c7..c2585ee 100644
--- a/lib/fiddle/truffleruby.rb
+++ b/lib/fiddle/truffleruby.rb
@@ -9,6 +9,9 @@
 # GNU General Public License version 2, or
 # GNU Lesser General Public License version 2.1.
 
+require_relative "truffleruby#{RUBY_ENGINE_VERSION.split('.', 2)[0]}"
+return
+
 module Truffle::FiddleBackend
 
   SIZEOF_CHAR   = Primitive.pointer_find_type_size(:char)
```

It's not difficult nor complex.

Or we can use #147 for TruffleRuby too instead of #148: #149

eregon:
> Or we can use #147 for TruffleRuby too instead of #148: #149

Yes, let's do that.
The relicense was faster than expected, great.

PR #147, opened 2024-09-28
Import JRuby implementation

kou:
Fix GH-104

lib/fiddle/jruby.rb is based on
https://github.com/jruby/jruby/blob/master/lib/ruby/stdlib/fiddle/jruby.rb .

Here are changes for it:

* Move `Fiddle::TYPE_*` to `Fiddle::Types::*`
* Add `Fiddle::Types::VARIADIC`
* Add `Fiddle::Types::CONST_STRING`
* Add `Fiddle::Types::BOOL`
* Add `Fiddle::Types::INT8_T`
* Add `Fiddle::Types::UINT8_T`
* Add `Fiddle::Types::INT16_T`
* Add `Fiddle::Types::UINT16_T`
* Add `Fiddle::Types::INT32_T`
* Add `Fiddle::Types::UINT32_T`
* Add `Fiddle::Types::INT64_T`
* Add `Fiddle::Types::UINT64_T`
* Add more `Fiddle::ALIGN_*` for the above new `Fiddle::Types::*`
* Add more `Fiddle::SIZEOF_*` for the above new `Fiddle::Types::*`
* Add support for specifying type as not only `Fiddle::Types::*` but also `Symbol` like `:int`
* Add support for variable size arguments in `Fiddle::Function`
* Add `Fiddle::Closure#free`
* Add `Fiddle::Closure#freed?`
* Add `Fiddle::Error` as base the error class
* Add `Fiddle::Pointer#call_free` and stop using `FFI::AutoPointer` in `Fiddle::Pointer`
* Add support for `Fiddle::Pointer.malloc {}` `Fiddle::Pointer`
* Add support for `Fiddle::Pointer#free=`
* Add `Fiddle::Pointer#freed?`
* Use binary string not C string for `Fiddle::Pointer#[]`
* Add `Fiddle::Handle.sym_defined?`
* Add `Fiddle::Handle#sym_defined?`
* Add `Fiddle::Handle::DEFAULT`
* Add `Fiddle::ClearedReferenceError`
* Add no-op `Fiddle::Pinned`

Some features are still "not implemented". So there are some "omit"s for JRuby in tests.

PR #148, opened 2024-09-29
Import TruffleRuby implementation

kou:
Fix GH-145

lib/fiddle/truffleruby.rb is based on
https://github.com/oracle/truffleruby/blob/master/lib/truffle/truffle/fiddle_backend.rb
.

Here are changes for it:

* Add `Fiddle::Types::VARIADIC`
* Add `Fiddle::Types::CONST_STRING`
* Add `Fiddle::Types::BOOL`
* Add `Fiddle::ALIGN_BOOL`
* Add `Fiddle::SIZEOF_BOOL`
* Add `Fiddle::SIZEOF_CONST_STRING`
* Add support for specifying type as not only `Fiddle::Types::*` but also
  `Symbol` like `:int`
* Add `Fiddle::Error` as base the error class
* Add support for `Fiddle::Pointer.malloc {}`
  `Fiddle::Pointer`
* Add support for `Fiddle.free(#to_int)`
* Accept `Fiddle::Function(need_gvl:)` but it's just ignored
* `Fiddle::Function#initialize`: Add an argument validation
* `Fiddle::Function#initialize`: Keep arguments as instance variables for
  getters
* Add support for `Fiddle::Handle.sym`
* Add support for `Fiddle::Handle.[]`
* Add support for `Fiddle::Handle.sym_defined?`
* Add support for `Fiddle::Handle#sym`
* Add support for `Fiddle::Handle#[]`
* Add support for `Fiddle::Handle#sym_defined?`
* Add support for `Fiddle::Pointer.malloc`
* Add support for `Fiddle::Pointer.to_ptr(#to_ptr)`
* Add support for `Fiddle::Pointer#free=`
* Add `Fiddle::Pointer#freed?`
* Add support for `Fiddle::Pointer#call_free`
* Add support for `Fiddle::Pointer#to_i`
* Add support for `Fiddle::Pointer#to_int`
* Add support for `Fiddle::Pointer#ptr`
* Add support for `Fiddle::Pointer#+@`
* Add support for `Fiddle::Pointer#ref`
* Add support for `Fiddle::Pointer#-@`
* Add support for `Fiddle::Pointer#null?`
* Add support for `Fiddle::Pointer#to_s`
* Add support for `Fiddle::Pointer#to_str`
* Add support for `Fiddle::Pointer#inspect`
* Add support for `Fiddle::Pointer#<=>`
* Add support for `Fiddle::Pointer#==`
* Add support for `Fiddle::Pointer#eql?`
* Add support for `Fiddle::Pointer#+`
* Add support for `Fiddle::Pointer#-`
* Add support for `Fiddle::Pointer#[]=`
* Add support for `Fiddle::Pointer#size`
* Add support for `Fiddle::Pointer#size=`
* Add `Fiddle::ClearedReferenceError`
* Add no-op `Fiddle::Pinned`
* Add `Fiddle::NULL`

Some features are still "not implemented". So there are some "omit"s
for TruffleRuby in tests.

### DIFF
--- a/lib/fiddle/truffleruby.rb
+++ b/lib/fiddle/truffleruby.rb
@@ -1,0 +1,623 @@
+# frozen_string_literal: true
+# truffleruby_primitives: true
+
+# Copyright (c) 2019, 2024 Oracle and/or its affiliates. All rights reserved. This
+# code is released under a tri EPL/GPL/LGPL license. You can use it,
+# redistribute it and/or modify it under the terms of the:
+#
+# Eclipse Public License version 2.0, or
+# GNU General Public License version 2, or
+# GNU Lesser General Public License version 2.1.
+
+module Truffle::FiddleBackend
+
+  SIZEOF_CHAR   = Primitive.pointer_find_type_size(:char)
+  SIZEOF_SHORT  = Primitive.pointer_find_type_size(:short)
+  SIZEOF_INT    = Primitive.pointer_find_type_size(:int)
+  SIZEOF_LONG   = Primitive.pointer_find_type_size(:long)
+
+  CHAR_NFI_TYPE  = "SINT#{SIZEOF_CHAR * 8}"
+  UCHAR_NFI_TYPE  = "UINT#{SIZEOF_CHAR * 8}"
+  SHORT_NFI_TYPE  = "SINT#{SIZEOF_SHORT * 8}"
+  USHORT_NFI_TYPE  = "UINT#{SIZEOF_SHORT * 8}"
+  INT_NFI_TYPE  = "SINT#{SIZEOF_INT * 8}"
+  UINT_NFI_TYPE  = "UINT#{SIZEOF_INT * 8}"
+  LONG_NFI_TYPE = "SINT#{SIZEOF_LONG * 8}"
+  ULONG_NFI_TYPE = "UINT#{SIZEOF_LONG * 8}"
+
+  SIZE_T_TYPEDEF = Truffle::Config.lookup('platform.typedef.size_t')
+
+  case SIZE_T_TYPEDEF
+  when 'ulong'
+    SIGNEDNESS_OF_SIZE_T = 1
+  else
+    raise NotImplementedError, "#{SIZE_T_TYPEDEF} not recognised"
+  end
+
+  def self.type_to_nfi(type)
+    if type.is_a?(Symbol)
+      type = Fiddle::Types.const_get(type.to_s.upcase)
+    end
+    if !type.is_a?(Integer) and type.respond_to?(:to_int)
+      type = type.to_int
+    end
+    case type
+    when Fiddle::TYPE_VOID
+      'VOID'
+    when Fiddle::TYPE_VOIDP, Fiddle::TYPE_CONST_STRING
+      'POINTER'
+    when Fiddle::TYPE_CHAR
+      CHAR_NFI_TYPE
+    when -Fiddle::TYPE_CHAR
+      UCHAR_NFI_TYPE
+    when Fiddle::TYPE_SHORT
+      SHORT_NFI_TYPE
+    when -Fiddle::TYPE_SHORT
+      USHORT_NFI_TYPE
+    when Fiddle::TYPE_INT
+      INT_NFI_TYPE
+    when -Fiddle::TYPE_INT
+      UINT_NFI_TYPE
+    when Fiddle::TYPE_LONG, Fiddle::TYPE_LONG_LONG
+      LONG_NFI_TYPE
+    when -Fiddle::TYPE_LONG, -Fiddle::TYPE_LONG_LONG
+      ULONG_NFI_TYPE
+    when Fiddle::TYPE_FLOAT
+      'FLOAT'
+    when Fiddle::TYPE_DOUBLE
+      'DOUBLE'
+    when Fiddle::TYPE_BOOL
+      'BOOL'
+    else
+      raise TypeError, "#{type} not implemented"
+    end
+  end
+
+  def self.type_from_config(type)
+    case type
+    when 'long'
+      Fiddle::Types::LONG
+    else
+      raise NotImplementedError, "integer #{type} not known"
+    end
+  end
+
+  def self.convert_ruby_to_native(type, val)
+    case type
+    when Fiddle::TYPE_CHAR
+      Integer(val)
+    when Fiddle::TYPE_VOIDP
+      get_pointer_value(val)
+    when Fiddle::TYPE_INT
+      Integer(val)
+    when -Fiddle::TYPE_LONG
+      Integer(val)
+    when Fiddle::TYPE_FLOAT, Fiddle::TYPE_DOUBLE
+      Float(val)
+    else
+      raise NotImplementedError, "#{val.inspect} to type #{type}"
+    end
+  end
+
+  def self.get_pointer_value(val)
+    if Primitive.is_a?(val, String)
+      # NOTE: Fiddle::TYPE_CONST_STRING wouldn't need inplace, but not defined yet by this file
+      Truffle::CExt.string_to_ffi_pointer_inplace(val)
+    elsif Primitive.is_a?(val, Fiddle::Pointer)
+      val.to_i
+    elsif val.respond_to?(:to_ptr)
+      val.to_ptr.to_i
+    elsif Primitive.nil?(val)
+      0
+    elsif Primitive.is_a?(val, Integer)
+      val
+    else
+      raise NotImplementedError, "#{val.inspect} to pointer"
+    end
+  end
+
+  def self.convert_native_to_ruby(type, val)
+    case type
+    when Fiddle::TYPE_VOID
+      nil
+    when Fiddle::TYPE_VOIDP
+      Fiddle::Pointer.new(Truffle::Interop.as_pointer(val))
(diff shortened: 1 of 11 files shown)


kou:
This is based on GH-147. So we should not merge this before GH-147.

eregon:
I changed the target branch so this PR shows the relevant diff.

lib/fiddle/truffleruby.rb
```diff
@@ -0,0 +1,623 @@
...
+# code is released under a tri EPL/GPL/LGPL license. You can use it,
+# redistribute it and/or modify it under the terms of the:
+#
+# Eclipse Public License version 2.0, or
+# GNU General Public License version 2, or
+# GNU Lesser General Public License version 2.1.
+
+module Truffle::FiddleBackend
+
+  SIZEOF_CHAR   = Primitive.pointer_find_type_size(:char)
```
eregon, line 14: `Primitive` might be renamed or changed, etc.
So far they were only used inside TruffleRuby so there was no compatibility needed whatsoever.
`Primitive`s so far have been an implementation detail (and in fact even the syntax to use them changed).

If a gem uses a Primitive then we need to make that Primitive stable.
One idea there is to annotate it in https://github.com/oracle/truffleruby (e.g. with `@Stable`, `stable = true`, `stable = "fiddle"` or `usedBy = "fiddle"`) so we know we shouldn't change its name or behavior.
And also make sure exposing that behavior as a Primitive makes sense or is necessary.

We should probably also expose a predicate to find out if a Primitive is defined or not, since the set of Primitives changes with different TruffleRuby versions. Maybe `TruffleRuby.primitive?(name)`.
That way if we want to use a new Primitive but it might not always be available it's possible.
And it's also nicer than comparing versions.

I don't want TruffleRuby to repeat the mistake of exposing all internal APIs as "public API because used in some gem" (which was the case for CRuby C API and for JRuby Java API).

lib/fiddle/truffleruby.rb
```diff
@@ -0,0 +1,623 @@
...
+  CHAR_NFI_TYPE  = "SINT#{SIZEOF_CHAR * 8}"
+  UCHAR_NFI_TYPE  = "UINT#{SIZEOF_CHAR * 8}"
+  SHORT_NFI_TYPE  = "SINT#{SIZEOF_SHORT * 8}"
+  USHORT_NFI_TYPE  = "UINT#{SIZEOF_SHORT * 8}"
+  INT_NFI_TYPE  = "SINT#{SIZEOF_INT * 8}"
+  UINT_NFI_TYPE  = "UINT#{SIZEOF_INT * 8}"
+  LONG_NFI_TYPE = "SINT#{SIZEOF_LONG * 8}"
+  ULONG_NFI_TYPE = "UINT#{SIZEOF_LONG * 8}"
+
+  SIZE_T_TYPEDEF = Truffle::Config.lookup('platform.typedef.size_t')
```
eregon, line 28: Everything under `Truffle::` is considered implementation details, with the exception of `Truffle::Interop`:
https://github.com/oracle/truffleruby/blob/master/doc/user/truffleruby-additions.md#unsupported-additional-functionality

It's a similar concern as with the Primitives, whatever is used in external gems needs to be made stable first.
Which might imply moving the method to maybe something like `TruffleRuby::FiddleSupport`, renaming the method or using a Primitive instead, etc.

lib/fiddle/truffleruby.rb
```diff
@@ -0,0 +1,623 @@
...
+
+# Copyright (c) 2019, 2024 Oracle and/or its affiliates. All rights reserved. This
+# code is released under a tri EPL/GPL/LGPL license. You can use it,
+# redistribute it and/or modify it under the terms of the:
+#
+# Eclipse Public License version 2.0, or
+# GNU General Public License version 2, or
+# GNU Lesser General Public License version 2.1.
+
+module Truffle::FiddleBackend
```
eregon, line 12: We shouldn't define `Truffle` or `Truffle::FiddleBackend` here but something like `Fiddle::TruffleRubyBackend`.

lib/fiddle/truffleruby.rb
eregon: One trouble with this is e.g. if this is changed/moved/renamed then it won't work on TruffleRuby 24.1.0 and older versions.
Then we would need another solution for older versions.

For example a stable `Primitive` seems better here and exposing `Truffle::Config.lookup`.

lib/fiddle/truffleruby.rb
```diff
@@ -0,0 +1,623 @@
+# frozen_string_literal: true
+# truffleruby_primitives: true
+
+# Copyright (c) 2019, 2024 Oracle and/or its affiliates. All rights reserved. This
+# code is released under a tri EPL/GPL/LGPL license. You can use it,
+# redistribute it and/or modify it under the terms of the:
+#
+# Eclipse Public License version 2.0, or
+# GNU General Public License version 2, or
+# GNU Lesser General Public License version 2.1.
```
eregon, line 10: We will need to relicense this code to upstream here (otherwise we would have to change the fiddle gem license).
From `git shortlog -sne lib/truffle/truffle/fiddle_backend.rb` it looks like it was only modified by Oracle employees, so then we should be able to relicense that, by relicensing https://github.com/oracle/truffleruby/blob/master/lib/truffle/truffle/fiddle_backend.rb under BSD-2-Clause.

kou: Thanks. It's very helpful.

lib/fiddle/truffleruby.rb
kou: You can change `lib/fiddle/truffleruby.rb` in this branch directly.

kou:
I've added `truflleruby-20` and `truffleruby-21` because:
* it seems that they are only available LTS based on https://www.graalvm.org/release-calendar/#previous-releases
* it seems that you're interested in old TruffleRuby releases

`ruby/setup-ruby` don't provide them for `macos-14` and `ubuntu-24.04`:

https://github.com/ruby/fiddle/actions/runs/11137969090/job/30952238981?pr=148#step:3:17
 
```text
Error: Error: Unavailable version 20.3.0 for truffleruby on macos-14
```
 
https://github.com/ruby/fiddle/actions/runs/11137969090/job/30952239103?pr=148#step:3:17
 
```text
Error: Error: Unavailable version 21.3.0 for truffleruby on macos-14
```

https://github.com/ruby/fiddle/actions/runs/11137969090/job/30952240601?pr=148#step:3:14

```text
Error: Error: Unavailable version 20.3.0 for truffleruby on ubuntu-24.04
```

https://github.com/ruby/fiddle/actions/runs/11137969090/job/30952240753?pr=148#step:3:14

```text
Error: Error: Unavailable version 21.3.0 for truffleruby on ubuntu-24.04
```

`truffleruby-20` on `macos-12` doesn't work because it doesn't work with recent Bundler:

https://github.com/ruby/fiddle/actions/runs/11137969090/job/30952235259?pr=148#step:3:27

```text
   Using latest Bundler for truffleruby-20.3.0 because the default Bundler gem is too old for that Ruby version
  Ruby 2.6-2.7 only works with Bundler 2.4
  /Users/runner/.rubies/truffleruby-20.3.0/bin/gem install bundler --force -v ~> 2.4.0
```

`truffleruby-21` on `macos-12` doesn't work because of internal Bundler error:

https://github.com/ruby/fiddle/actions/runs/11137969090/job/30952235963?pr=148#step:4:10

```text
NameError: method `search_by_dependency' not defined in Bundler::Index
```

Should we really care about old TruffleRuby releases?

lib/fiddle/truffleruby.rb
kou: We can release a new version for newer TruffleRuby if it's needed.
Old TruffleRuby users can pin Fiddle gem to old version.

lib/fiddle/truffleruby.rb
kou: We can do it or we can use `ffi` API like JRuby does.

eregon:
> I've added truffleruby-20 and truffleruby-21 because:

The versions at https://www.graalvm.org/release-calendar/#previous-releases are very confusing unfortunately, as they use Java versions, while Truffle/TruffleRuby versions use the year.
So truffleruby-20 and truffleruby-21 are ancient, from 2020 and 2021.

The current GraalVM LTS is GraalVM for JDK 21, which corresponds to TruffleRuby 23.1.0.
So it might make sense to test that and/or 24.0 as well.

I think we should solve it for all existing TruffleRuby releases if possible and #146 is one way to do that.
This PR is unlikely to work on older releases, and might not even work on the latest release, once we avoid depending on internal & changing APIs.

lib/fiddle/truffleruby.rb
eregon: > Old TruffleRuby users can pin Fiddle gem to old version.

The error for end users would be very unclear though (e.g. some NoMethodError), and they would likely have no idea how to solve this from the error message.
Also if for whatever reason it's a program run without Bundler then I don't think there is a way to pin the Fiddle version.

That's why I have always thought and still think the best solution is this PR for upcoming TruffleRuby versions, and using the stdlib for older TruffleRuby versions.
This is also what the FFI gem does for example:
https://github.com/ffi/ffi/blob/c128cede750242fe19945af8bd6c797728489ad5/lib/ffi.rb#L14-L27

kou:
We use #149 instead of this.

PR #149, opened 2024-10-07
Use JRuby implementation for TruffleRuby

kou:
Fix GH-145

Rename `lib/fiddle/jruby.rb` to `lib/fiddle/ffi_backend.rb` as a generic ffi gem API based implementation.
JRuby and TruffleRuby use `lib/fiddle/ffi_backend.rb`.


lib/fiddle/jruby.rb
```diff
@@ -158,6 +158,17 @@ def call(*args, &block);
...
+        args.size.times do |i|
+          arg = args[i]
+          next unless arg.respond_to?(:to_ptr)
+          loop do
+            arg = arg.to_ptr
+            break if arg.is_a?(FFI::Pointer)
+            break unless arg.respond_to?(:to_ptr)
+          end
+          args[i] = arg
+        end
```
eregon, line 171: Interesting that this is needed on TruffleRuby but apparently not on JRuby.
I'll repro to see what's the difference there.

test/fiddle/test_closure.rb
```diff
@@ -6,6 +6,12 @@
 
 module Fiddle
   class TestClosure < Fiddle::TestCase
+    def setup
+      if RUBY_ENGINE == "truffleruby"
+        omit("FFI::Function don't accept #call-able object with TruffleRuby")
```
eregon, line 11: I'll look into this, TruffleRuby does pass most FFI gem tests so it's surprising, maybe this is not well tested or is not an officially supported API.

eregon (commented):
This looks great, I'd like to investigate a couple things.

lib/fiddle/jruby.rb
```diff
@@ -158,6 +158,17 @@ def call(*args, &block);
           end
```
eregon, line 158: I think it would be good to rename this file to `ffi_backend.rb` or so and avoid a JRuby namespace since it is effectively independent from JRuby but it doesn't like that currently due to file naming and things like `Fiddle::JRuby`.
It's helpful to have a small diff at this point though, so I think that's best done later, e.g. a separate PR.

kou: I think that we don't need to do it because this is a short-term workaround for TruffleRuby.
This file will be used by only JRuby eventually.

lib/fiddle/jruby.rb
eregon: OK it's needed, details explained in https://github.com/ruby/fiddle/pull/149/commits/0631d46cddb4dc64ff449ddc62524c6767e9b66b

test/fiddle/test_closure.rb
eregon: Fixed in https://github.com/ruby/fiddle/pull/149/commits/41c21f5887af6133ed5ee347ac6143cbc396c382 so all such tests pass now

lib/fiddle/jruby.rb
eregon: Since the FFI backend seems to work well I think we should just use it long term, for both JRuby and TruffleRuby.

test/fiddle/test_func.rb
```diff
@@ -78,16 +78,20 @@ def call(x, y)
...
         bug4929 = '[ruby-core:37395]'
-        buff = "9341"
+        buff = +"9341"
         under_gc_stress do
           qsort.call(buff, buff.size, 1, callback)
         end
         assert_equal("1349", buff, bug4929)
+
+        # Ensure the test didn't mutate String literals
+        assert_equal("93" + "41", untouched)
```
eregon, line 94: Without the added `+` above this was failing, which meant Fiddle tests were permanently changing the meaning of all `"9341"` string literals.

Interestingly this sanity check actually fails on JRuby, I'll file an issue there.

eregon (approved):
I think this is ready, I am happy with this solution.

lib/fiddle/truffleruby.rb
kou, line 1: Can we remove this file?

eregon: Done

test/fiddle/test_func.rb
headius: The JRuby issue has been fixed on master and will be in JRuby 9.4.9.0.

We won't be updating 9.4 to use the fiddle gem in any case, but the test could be un-omitted now.

headius:
@kou Thanks so much for the extra work on this!

PR #156, opened 2024-10-18
Fix Fiddle::Handle.new for a missing library in the FFI backend

eregon:
* From https://github.com/ruby/reline/issues/766#issuecomment-2422135968

cc @headius

### DIFF
--- a/lib/fiddle/ffi_backend.rb
+++ b/lib/fiddle/ffi_backend.rb
@@ -475,8 +475,11 @@ module Fiddle
     RTLD_NOW = FFI::DynamicLibrary::RTLD_NOW
 
     def initialize(libname = nil, flags = RTLD_LAZY | RTLD_GLOBAL)
-      @lib = FFI::DynamicLibrary.open(libname, flags) rescue LoadError
-      raise DLError.new("Could not open #{libname}") unless @lib
+      begin
+        @lib = FFI::DynamicLibrary.open(libname, flags)
+      rescue LoadError, RuntimeError # LoadError for JRuby, RuntimeError for TruffleRuby
+        raise DLError, "Could not open #{libname}"
+      end
 
       @open = true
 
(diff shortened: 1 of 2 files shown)


lib/fiddle/ffi_backend.rb
```diff
@@ -475,8 +475,12 @@ class Handle
     RTLD_NOW = FFI::DynamicLibrary::RTLD_NOW
 
     def initialize(libname = nil, flags = RTLD_LAZY | RTLD_GLOBAL)
-      @lib = FFI::DynamicLibrary.open(libname, flags) rescue LoadError
-      raise DLError.new("Could not open #{libname}") unless @lib
+      @lib =
+        begin
+          FFI::DynamicLibrary.open(libname, flags)
+        rescue
```
eregon, line 481: Plain `rescue` because it's LoadError on JRuby and RuntimeError for TruffleRuby.

lib/fiddle/ffi_backend.rb
```diff
@@ -475,8 +475,11 @@ class Handle
     RTLD_NOW = FFI::DynamicLibrary::RTLD_NOW
 
     def initialize(libname = nil, flags = RTLD_LAZY | RTLD_GLOBAL)
-      @lib = FFI::DynamicLibrary.open(libname, flags) rescue LoadError
-      raise DLError.new("Could not open #{libname}") unless @lib
+      begin
+        @lib = FFI::DynamicLibrary.open(libname, flags)
+      rescue
```
eregon, line 480: Plain `rescue` because it's LoadError on JRuby and RuntimeError for TruffleRuby.

kou:
Good catch.

headius:
Thank you for the fix. I have confirmed it locally using fiddle 1.1.4 and JRuby 10. The reline issue is gone.

Issue #168, opened 2025-01-03
bool struct member parsing not working 


wallacerunner:
This script
```ruby
require 'fiddle/import'

module Test
  extend Fiddle::Importer

  test_struct = struct [
    'int int_member',
    'bool boolean_member'
  ]
end
```
fails with this error:
```
PS D:\code\ruby\sdl_test> ruby .\struct_test.rb
C:/Users/Dee/scoop/apps/ruby/3.3.6-2/lib/ruby/3.3.0/fiddle/cparser.rb:261:in `parse_ctype': unknown type: bool (Fiddle::DLError)
        from C:/Users/Dee/scoop/apps/ruby/3.3.6-2/lib/ruby/3.3.0/fiddle/cparser.rb:71:in `block in parse_struct_signature'
        from C:/Users/Dee/scoop/apps/ruby/3.3.6-2/lib/ruby/3.3.0/fiddle/cparser.rb:43:in `each'
        from C:/Users/Dee/scoop/apps/ruby/3.3.6-2/lib/ruby/3.3.0/fiddle/cparser.rb:43:in `parse_struct_signature'
        from C:/Users/Dee/scoop/apps/ruby/3.3.6-2/lib/ruby/3.3.0/fiddle/import.rb:223:in `struct'
        from ./struct_test.rb:6:in `<module:Test>'
        from ./struct_test.rb:3:in `<main>'
```

Looks like `parse_struct_signature` passes the whole `bool boolean_member` string to `parse_ctype`, where it fails to be parsed because case expression tries to match on `"bool"` (i.e. a plain string) instead of a regex, something like `/\Abool(?:\s+\w+)?\z/`:
https://github.com/ruby/fiddle/blob/2131aa1f504ab7357f600cb8fb2efbf43c8cf6ce/lib/fiddle/cparser.rb#L250

kou:
Could you open a PR with a test?

PR #169, opened 2025-01-06
add regex for bool parsing & test struct w/ bool parsing

wallacerunner:
GitHub: fix GH-168

Struct parsing invokes "parse_ctype" on the whole member signature, which fails if member type is "bool" due to plain string matching for it. This change updates "bool" type matching to a regexp, so TYPE_BOOL is correctly parsed for a whole signature like "bool toggle" as well as just "bool".

### DIFF
--- a/test/fiddle/test_cparser.rb
+++ b/test/fiddle/test_cparser.rb
@@ -277,6 +277,10 @@ module Fiddle
       assert_equal [[TYPE_INT,TYPE_VOIDP,TYPE_VOIDP], ['x', 'cb', 'name']], parse_struct_signature('int x; void (*cb)(); const char* name')
     end
 
+    def test_struct_bool
+      assert_equal [[TYPE_INT, TYPE_BOOL], ['x', 'toggle']], parse_struct_signature('int x; bool toggle')
+    end
+
     def test_struct_undefined
       assert_raise(DLError) { parse_struct_signature(['int i', 'DWORD cb']) }
     end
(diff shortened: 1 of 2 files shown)


test/fiddle/test_cparser.rb
```diff
@@ -277,6 +277,10 @@ def test_struct_string
       assert_equal [[TYPE_INT,TYPE_VOIDP,TYPE_VOIDP], ['x', 'cb', 'name']], parse_struct_signature('int x; void (*cb)(); const char* name')
     end
 
+    def test_struct_bool
+      assert_equal [[TYPE_INT, TYPE_BOOL], ['x', 'toggle']], parse_struct_signature('int x; bool toggle')
```
kou, line 281: Could you ensure using `(...)` for newly added assertions to avoid warnings?
See also:  https://bugs.ruby-lang.org/issues/20922

```suggestion
      assert_equal([[TYPE_INT, TYPE_BOOL], ['x', 'toggle']],
                   parse_struct_signature('int x; bool toggle'))
```

kou (approved):
Could you explain what is the problem and how to solve it in the PR description?
We'll use the PR description for commit message.

wallacerunner:
Now that I gave it some thought, shouldn't this be the other way around? I.e. isn't parse_struct supposed to invoke parse_ctype on just the type definition, and not parse_ctype running regexp on given string? 
Still, I did not look into it deeply, and this PR seems like a much easier solution.

kou:
We can revisit it as a separated task.
Could you open a new issue for it?

PR #170, opened 2025-01-23
Update ci.yml

MSP-Greg:
Saw https://github.com/ruby/ruby/pull/12616, and ended up here.

Updates several small items in the main CI file.


kou:
Thanks.

PR #171, opened 2025-01-30
Fix test dlopen for libc.so filename with more than one digit

lucaskanashiro:
In Debian alpha architecture, we have libc.so.6.1 which causes a failure in this test:

     1) Failure:
    TestFiddle#test_dlopen_linker_script_group_linux [/<<PKGBUILDDIR>>/test/fiddle/test_fiddle.rb:52]:
    <"libc.so"> expected but was
    <"libc.so.6">.

Bug-Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1085238

### DIFF
--- a/test/fiddle/test_fiddle.rb
+++ b/test/fiddle/test_fiddle.rb
@@ -83,7 +83,7 @@ class TestFiddle < Fiddle::TestCase
     handle = Fiddle.dlopen("libc.so")
     begin
       assert_equal("libc.so",
-                   File.basename(handle.file_name, ".*"))
+                   File.basename(handle.file_name, ".*")).gsub(/\.so.\d+/,'.so'))
     ensure
       handle.close
     end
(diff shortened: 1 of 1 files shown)


test/fiddle/test_fiddle.rb
```diff
@@ -83,7 +83,7 @@ def test_dlopen_linker_script_group_linux
     handle = Fiddle.dlopen("libc.so")
     begin
       assert_equal("libc.so",
-                   File.basename(handle.file_name, ".*"))
+                   File.basename(handle.file_name, ".*")).gsub(/\.so.\d+/,'.so'))
```
kou, line 86: Let's use `gsub` for both of `libc.so.6` and `libc.so.6.1`:

```suggestion
                   File.basename(handle.file_name).gsub(/\.so(\.\d+)+\z/, ".so"))
```

PR #172, opened 2025-01-31
Define Fiddle.last_error family and Fiddle.dlopen statically

nobu:
`RUBY_ENGINE` and `Fiddle::WINDOWS` should not change in a process, no need to be checked inside the methods.

Also, `win32_last_error` and `win32_last_socket_error` are equal to `last_error` on JRuby.

### DIFF
--- a/lib/fiddle.rb
+++ b/lib/fiddle.rb
@@ -10,98 +10,72 @@ require 'fiddle/function'
 require 'fiddle/version'
 
 module Fiddle
-  if WINDOWS
-    # Returns the last win32 +Error+ of the current executing +Thread+ or nil
-    # if none
-    def self.win32_last_error
-      if RUBY_ENGINE == 'jruby'
-        errno = FFI.errno
-        errno == 0 ? nil : errno
-      else
+  case RUBY_ENGINE
+  when 'jruby'
+    def self.last_error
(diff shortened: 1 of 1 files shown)


lib/fiddle.rb
```diff
@@ -10,98 +10,72 @@
...
+        alias win32_last_error= last_error=
+        alias win32_last_socket_error last_error
+        alias win32_last_socket_error= last_error=
       end
     end
-  end
-
-  # Returns the last +Error+ of the current executing +Thread+ or nil if none
-  def self.last_error
-    if RUBY_ENGINE == 'jruby'
```
eregon, line 58: Mmh, actually `FFI.last_error{,=}` should also use FFI.errno on TruffleRuby, it seems somehow I missed this, maybe because there are no tests for this?

Orthogonal to this PR, so feel free to merge this first.

eregon: Ah it wasn't caught because of https://github.com/ruby/fiddle/blob/d5fd917123ca3cff9a976a4505848589cf822e49/test/fiddle/test_function.rb#L128-L131

eregon: => https://github.com/ruby/fiddle/pull/173

PR #173, opened 2025-01-31
Fix Fiddle.last_error on FFI backend and improve test to work for all

eregon:


test/fiddle/test_function.rb
```diff
@@ -126,14 +126,16 @@ def call one
...
 
     def test_last_error
-      if ffi_backend?
-        omit("Fiddle.last_error doesn't work with FFI backend")
-      end
+      require 'rbconfig/sizeof'
+
+      strtol = Function.new(@libc['strtol'], [TYPE_VOIDP, TYPE_VOIDP, TYPE_INT], TYPE_LONG)
+
+      assert_nil Fiddle.last_error unless ffi_backend?
```
eregon, line 133: FWIW this only works on CRuby due to the `Fiddle.last_error = nil` in `setup`.
On the FFI backend there is no clean way to do this, and there seems to be little value to return `nil` on `Fiddle.last_error`, as that can only happen if:
* there never was a call using Fiddle yet
* the user called `Fiddle.last_error = nil` explicitly, which seems not really useful

Maybe setup should `Fiddle.last_error = 0` instead and then we could test that here? Let me know what you prefer.

eregon:
Mmh, JRuby on Windows seems to have a bug here, cc @headius:
https://github.com/ruby/fiddle/actions/runs/13081426763/job/36505655327?pr=173#step:7:31
```
assert_equal RbConfig::LIMITS["LONG_MAX"], strtol.call((2**128).to_s, nil, 10)
<9223372036854775807> expected but was
<2147483647>
```
LONG_MAX on Windows is 2147483647 ([LLP64](https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models)), but `RbConfig::LIMITS["LONG_MAX"]` on JRuby on Windows reports 9223372036854775807 (incorrect).
(I made sure the value before `expected` is the LHS, so it's not the other way around)

test/fiddle/test_function.rb
kou: Let's use `Fiddle.last_error = 0`.

kou:
https://github.com/jruby/jruby/issues/8357 may be related for the JRuby behavior.

test/fiddle/test_function.rb
eregon: Done

eregon:
Worked around the JRuby on Windows failure by not asserting the result of stroll(), it's not really important, we are testing errno.

eregon:
There is another error for JRuby on Windows, I'll skip that test for it then

eregon:
CI should be green now and this PR should be ready

lib/fiddle.rb
```diff
@@ -22,9 +21,11 @@ def self.last_error= error
 
     if WINDOWS
       class << self
-        alias win32_last_error last_error
+        def win32_last_error
+          FFI.errno.nonzero?
```
kou, line 25:
```suggestion
          last_error.nonzero?
```

eregon: I think it's better as-is to decouple each other and no accidentally break the other e.g. when changing `last_error`.
But if you prefer that feel free to apply it.

Maybe `win32_last_error`/`win32_last_socket_error` should never return `nil` either, but I don't know enough about Windows errors to make that call.

kou: If we want to decouple `last_error` and `win32_last_error`, we should stop `alias win32_last_error= last_error=` too.

I think that it's an overengineering without any real world problem but I'm OK with either of them:

```ruby
def win32_last_error
  last_error.nonzero?
end
alias win32_last_error= last_error=
```

```ruby
def win32_last_error
  FFI.errno.nonzero?
end
def win32_last_error=(errno)
  FFI.errno = error || 0
end
```

eregon: OK, I did the second one (since they are supposed to be independent things but are not currently on JRuby).

kou:
Thanks.

PR #176, opened 2025-04-19
Make `Fiddle::Pointer#to_s` with JRuby consistent with CRuby

ankane:
Hi, this PR makes the behavior of `Fiddle::Pointer#to_s` consistent between CRuby and JRuby.

```ruby
ptr = Fiddle::Pointer.malloc(4)
p ptr.to_s(4)
```

currently outputs `"\x00\x00\x00\x00"` with CRuby and `""` with JRuby.

### DIFF
--- a/lib/fiddle/ffi_backend.rb
+++ b/lib/fiddle/ffi_backend.rb
@@ -408,10 +408,9 @@ module Fiddle
     end
     alias to_int to_i
 
-    # without \0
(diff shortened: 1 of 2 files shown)


kou:
Good catch!

ankane:
Thanks. Just fyi, the test failures are due to https://github.com/jruby/jruby/issues/8780.

PR #180, opened 2025-04-21
Fix memory corruption with FFI backend

ankane:
To reproduce with JRuby or TruffleRuby:

```ruby
require "fiddle"

1000.times do
  Fiddle::Pointer[rand(255).chr*16]
end
```

`put_string` adds a null byte at the end vs `write_bytes`. This is not caught by the bounds check since the underlying `FFI::Pointer` size is not set.

### DIFF
--- a/lib/fiddle/ffi_backend.rb
+++ b/lib/fiddle/ffi_backend.rb
@@ -240,7 +240,7 @@ module Fiddle
     def self.to_ptr(value)
       if value.is_a?(String)
         cptr = Pointer.malloc(value.bytesize)
-        cptr.ffi_ptr.put_string(0, value)
+        cptr.ffi_ptr.put_bytes(0, value)
         cptr
 
       elsif value.is_a?(Array)
--- a/lib/fiddle/ffi_backend.rb
+++ b/lib/fiddle/ffi_backend.rb
@@ -412,7 +412,7 @@ module Fiddle
       if len
         ffi_ptr.read_string(len)
       else
-        ffi_ptr.get_string(0)
+        ffi_ptr.get_string(0, @size)
       end
     rescue FFI::NullPointerError
       raise DLError.new("NULL pointer access")


lib/fiddle/ffi_backend.rb
```diff
@@ -239,7 +239,7 @@ def self.from_native(value, ctx)
 
     def self.to_ptr(value)
       if value.is_a?(String)
-        cptr = Pointer.malloc(value.bytesize)
+        cptr = Pointer.malloc(value.bytesize + 1)
         cptr.ffi_ptr.put_string(0, value)
```
kou, line 243: We may need to use `put_bytes` instead...

```suggestion
        cptr = Pointer.malloc(value.bytesize)
        cptr.ffi_ptr.put_bytes(0, value)
```

ankane: Tried that initially with `write_bytes`, but looks like some of the tests expect an additional null byte.

ankane: Nevermind, looks like the failing tests are due to an issue with the `to_s` method.

kou:
Thanks.

ankane:
Thank you

lib/fiddle/ffi_backend.rb
eregon: @kou Is it expected though that `Fiddle::Pointer.to_ptr("abc")` doesn't guarantee to return a \0-terminated native string?

eregon: This is what the C extension does:
https://github.com/ruby/fiddle/blob/7a4c6daffa3ed148e02a98ebcd3e0f1009a37393/ext/fiddle/pointer.c#L784-L787

StringValuePtr AFAIK always returns a \0 terminated char* in CRuby.
So then
```ruby
cptr = Pointer.malloc(value.bytesize + 1)
cptr.ffi_ptr.put_string(0, value)
```
is I believe the correct way.

i.e. it was correct as of [07a606f](https://github.com/ruby/fiddle/commit/07a606f01c626fbf1322cdfbef6cd8027e19b111)

eregon: @ankane Could you make a PR going back to that and also add a test that `Fiddle::Pointer.to_ptr("abc")[3]` is always 0?

Also maybe the Pointer#size should be checked to be 4 and not 3, and then https://github.com/ruby/fiddle/blob/7a4c6daffa3ed148e02a98ebcd3e0f1009a37393/ext/fiddle/pointer.c#L787 adapted to + 1.

ankane: Made the behavior consistent with CRuby in #187.

PR #181, opened 2025-04-21
Fix pointer class for closure arguments for FFI backend

ankane:
Currently, pointer arguments for closures use the FFI pointer class instead of the Fiddle pointer class with the FFI backend.

### DIFF
--- a/lib/fiddle/ffi_backend.rb
+++ b/lib/fiddle/ffi_backend.rb
@@ -188,8 +188,10 @@ module Fiddle
       end
       return_type = Fiddle::FFIBackend.to_ffi_type(@ctype)
       raise "#{self.class} must implement #call" unless respond_to?(:call)
-      callable = method(:call)
-      @function = FFI::Function.new(return_type, ffi_args, callable, convention: abi)
+      wrapper = lambda do |*args|
+        call(*args.map { |v| v.is_a?(FFI::Pointer) ? Pointer.new(v) : v })
+      end
+      @function = FFI::Function.new(return_type, ffi_args, wrapper, convention: abi)
       @freed = false
     end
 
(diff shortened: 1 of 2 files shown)


test/fiddle/test_closure.rb
```diff
@@ -87,6 +87,18 @@ def call(bool)
...
 
+    def test_pointer
+      closure_class = Class.new(Closure) do
+        def call(ptr)
+          ptr.is_a?(Pointer)
+        end
+      end
+      closure_class.create(:bool, [:voidp]) do |closure|
+        func = Function.new(closure, [:voidp], :bool)
+        assert_equal(true, func.call(Pointer["hello"]))
```
kou, line 98: Could you use power assert style for boolean?

```suggestion
        assert do
          func.call(Pointer["hello"])
        end
```

lib/fiddle/ffi_backend.rb
```diff
@@ -189,7 +189,10 @@ def initialize(ret, args, abi = Function::DEFAULT)
       return_type = Fiddle::FFIBackend.to_ffi_type(@ctype)
       raise "#{self.class} must implement #call" unless respond_to?(:call)
       callable = method(:call)
-      @function = FFI::Function.new(return_type, ffi_args, callable, convention: abi)
+      wrapper = lambda do |*args|
+        callable.(*args.map { |v| v.is_a?(FFI::Pointer) ? Pointer.new(v) : v })
```
kou, line 193: Can we use `call` directly here?

```suggestion
        call(*args.map { |v| v.is_a?(FFI::Pointer) ? Pointer.new(v) : v })
```

kou:
Thanks.

ankane:
Thanks again

PR #184, opened 2025-04-22
Make `to_str` consistent between backends for negative pointer sizes

ankane:
Also added more tests for the current behavior.

```ruby
require "fiddle"

ptr = Fiddle::Pointer["hello\0world"]
ptr.size = 0

p (ptr + 1).size
p (ptr + 1).to_s(5)
p (ptr + 1).to_s
p (ptr + 1).to_str(5)
p (ptr + 1).to_str
```

Output

```text
-1
"ello\x00"
"ello"
"ello\x00"
repro.rb:10:in 'Fiddle::Pointer#to_str': negative string size (or size too big) (ArgumentError)
```

### DIFF
--- a/lib/fiddle/ffi_backend.rb
+++ b/lib/fiddle/ffi_backend.rb
@@ -432,6 +432,9 @@ module Fiddle
       if len
         ffi_ptr.read_string(len)
       else
+        if @size < 0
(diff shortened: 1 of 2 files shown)


kou:
Thanks.

ankane:
Thanks

PR #185, opened 2025-04-22
Fix `Fiddle::Pointer#ref` behavior for FFI backend

### LINKED ISSUE #182: Difference with `Fiddle::Pointer#ref` and FFI backend
Hi, with the C backend, the code below updates `ptr` in-place. However, with the FFI backend, `ptr` is unchanged (with both JRuby and TruffleRuby).

```ruby
require "fiddle"

ptr = Fiddle::Pointer["hello"]
i = ptr.to_i

ptr.ref[0, Fiddle::SIZEOF_VOIDP] = Fiddle::Pointer["world"]
p ptr.to_i == i # false for C backend, true for FFI backend
```

I'm having trouble finding the cause (to try to make the behavior consistent), so wanted to see if anyone had ideas.

ankane:
Fixes #182 by making the pointer address dynamic.

```ruby
require "fiddle"

ptr = Fiddle::Pointer["hello"]
ptr.ref[0, Fiddle::SIZEOF_VOIDP] = Fiddle::Pointer["world"].ref
p ptr.to_s
```

The code above previously returned `"hello"` with the FFI backend.

### DIFF
--- a/test/fiddle/test_pointer.rb
+++ b/test/fiddle/test_pointer.rb
@@ -187,6 +187,19 @@ module Fiddle
       assert_nil(ptr <=> 10, '10 should not be comparable')
     end
 
+    def test_ref
+      ptr = Fiddle::Pointer["hello"]
+      ref = ptr.ref
+      assert_equal(0, ref.size)
+      assert_nil(ref.free)
+      assert_equal(ptr, ref.ptr)
+
+      ptr2 = Fiddle::Pointer["world"]
+      ptr.ref[0, Fiddle::SIZEOF_VOIDP] = ptr2.ref
+      assert_equal("world", ptr.to_s)
+      assert_equal(ptr.to_i, ptr2.to_i)
+    end
+
     def test_ref_ptr
       if ffi_backend?
         omit("Fiddle.dlwrap([]) isn't supported with FFI backend")
--- a/lib/fiddle/ffi_backend.rb
+++ b/lib/fiddle/ffi_backend.rb
@@ -219,7 +219,6 @@ module Fiddle
   class ClearedReferenceError < Error; end
 
   class Pointer
-    attr_reader :ffi_ptr
     extend FFI::DataConverter
     native_type FFI::Type::Builtin::POINTER
 
@@ -285,7 +284,7 @@ module Fiddle
           value = value.to_str(args[1])
         end
 
-        @ffi_ptr.put_bytes(args[0], value, 0, args[1])
+        ffi_ptr.put_bytes(args[0], value, 0, args[1])
       elsif args.size == 1
         if value.is_a?(Fiddle::Pointer)
           value = value.to_str(args[0] + 1)
@@ -293,7 +292,7 @@ module Fiddle
           value = value.chr
         end
 
-        @ffi_ptr.put_bytes(args[0], value, 0, 1)
+        ffi_ptr.put_bytes(args[0], value, 0, 1)
       end
     rescue FFI::NullPointerError
       raise DLError.new("NULL pointer access")
@@ -332,7 +331,12 @@ module Fiddle
       end
       @free = free
       @ffi_ptr = ptr
+      @addr_ptr = nil
       @freed = false
+    end
+
+    def ffi_ptr
+      @addr_ptr ? @addr_ptr.get_pointer(0) : @ffi_ptr
     end
 
     module LibC
@@ -363,11 +367,11 @@ module Fiddle
     end
 
     def null?
-      @ffi_ptr.null?
+      ffi_ptr.null?
     end
 
     def to_ptr
-      @ffi_ptr
+      ffi_ptr
     end
 
     def size
@@ -386,9 +390,9 @@ module Fiddle
       return if @free.nil?
       return if @freed
       if @free == RUBY_FREE
-        LibC::FREE.call(@ffi_ptr)
+        LibC::FREE.call(ffi_ptr)
       else
-        @free.call(@ffi_ptr)
+        @free.call(ffi_ptr)
       end
       @freed = true
     end
@@ -492,10 +496,8 @@ module Fiddle
     end
 
     def ref
-      cptr = Pointer.malloc(FFI::Type::POINTER.size, RUBY_FREE)
-      cptr.ffi_ptr.put_pointer(0, ffi_ptr)
-      cptr.size = 0
-      cptr
+      @addr_ptr ||= FFI::MemoryPointer.new(:pointer).put_pointer(0, @ffi_ptr)
+      Pointer.new(@addr_ptr, 0)
     end
   end
 
--- a/lib/fiddle/ffi_backend.rb
+++ b/lib/fiddle/ffi_backend.rb
@@ -218,6 +218,7 @@ module Fiddle
   class DLError < Error; end
   class ClearedReferenceError < Error; end
 
+  # Pointer isn't thread safe for now
   class Pointer
     extend FFI::DataConverter
     native_type FFI::Type::Builtin::POINTER


ankane:
Sorry for the noise. I think a better test would be:

```ruby
ptr2 = Fiddle::Pointer["world"]
ptr.ref[0, Fiddle::SIZEOF_VOIDP] = ptr2.ref
assert_equal "world", ptr.to_s
assert_equal ptr.to_i, ptr2.to_i
```

But will hold off on updating again until after initial feedback.

test/fiddle/test_pointer.rb
```diff
@@ -187,6 +187,19 @@ def test_cmp
       assert_nil(ptr <=> 10, '10 should not be comparable')
     end
 
+    def test_ref
+      ptr = Fiddle::Pointer["hello"]
+      ref = ptr.ref
+      assert_equal 0, ref.size
+      assert_nil ref.free
+      assert_equal ptr, ref.ptr
```
kou, line 195: Could you use parenthesis for assertions?
See also: https://bugs.ruby-lang.org/issues/20922

```suggestion
      assert_equal(0, ref.size)
      assert_nil(ref.free)
      assert_equal(ptr, ref.ptr)
```

test/fiddle/test_pointer.rb
```diff
@@ -187,6 +187,19 @@ def test_cmp
...
+
+      ptr2 = Fiddle::Pointer["world"]
+      ptr.ref[0, Fiddle::SIZEOF_VOIDP] = ptr2.ref
+      assert_equal "world", ptr.to_s
+      assert_equal ptr.to_i, ptr2.to_i
```
kou, line 200:
```suggestion
      assert_equal("world", ptr.to_s)
      assert_equal(ptr.to_i, ptr2.to_i)
```

kou (commented):
Thanks.

I hadn't thought of this approach!

ankane:
Updated the assertions. One thing I'm not sure about is if it's better to create the address pointer in `initialize` or lazily in `ref` (and if lazily, does there need to be any synchronization to ensure only one address pointer is created for the object).

kou:
Good point.

In general, we should use the lazy approach because `#ref` isn't used for all use cases.

The current `Pointer` implementation isn't thread safe. For example, `#call_free` is not thread safe.

If you don't have any use case that it's better that `Pointer` is thread safe for now, how about adding a comment that "Pointer isn't thread safe for now" for now?
Or it may be better that a `Pointer` that is returned by `#ref` refers the base `Pointer` something like:

```diff
diff --git a/lib/fiddle/ffi_backend.rb b/lib/fiddle/ffi_backend.rb
index f6b3d12..025004d 100644
--- a/lib/fiddle/ffi_backend.rb
+++ b/lib/fiddle/ffi_backend.rb
@@ -220,6 +220,8 @@ module Fiddle
 
   class Pointer
     attr_reader :ffi_ptr
+    protected attr_writer :ffi_ptr
+    protected attr_writer :resolved_pointer
     extend FFI::DataConverter
     native_type FFI::Type::Builtin::POINTER
 
@@ -282,12 +284,14 @@ module Fiddle
           value = self.class.new(value)
         end
         if value.is_a?(Fiddle::Pointer)
+          @resolved_pointer.ffi_ptr = value.ffi_ptr if @resolved_pointer
           value = value.to_str(args[1])
         end
 
         @ffi_ptr.put_bytes(args[0], value, 0, args[1])
       elsif args.size == 1
         if value.is_a?(Fiddle::Pointer)
+          @resolved_pointer.ffi_ptr = value.ffi_ptr if @resolved_pointer
           value = value.to_str(args[0] + 1)
         else
           value = value.chr
@@ -332,6 +336,7 @@ module Fiddle
       end
       @free = free
       @ffi_ptr = ptr
+      @resolved_pointer = nil
       @freed = false
     end
 
@@ -497,6 +502,7 @@ module Fiddle
     def ref
       cptr = Pointer.malloc(FFI::Type::POINTER.size, RUBY_FREE)
       cptr.ffi_ptr.put_pointer(0, ffi_ptr)
+      cptr.resolved_pointer = self
       cptr.size = 0
       cptr
     end
```

ankane:
I tried applying the patch, but the tests no longer pass (unless I messed something up). Added a comment to the `Pointer` class in the commit above.

kou:
Ah, we needed more changes with the approach:

```diff
diff --git a/lib/fiddle/ffi_backend.rb b/lib/fiddle/ffi_backend.rb
index f6b3d12..6c19d75 100644
--- a/lib/fiddle/ffi_backend.rb
+++ b/lib/fiddle/ffi_backend.rb
@@ -220,6 +220,8 @@ module Fiddle
 
   class Pointer
     attr_reader :ffi_ptr
+    protected attr_writer :ffi_ptr
+    protected attr_accessor :resolved_pointer
     extend FFI::DataConverter
     native_type FFI::Type::Builtin::POINTER
 
@@ -282,12 +284,26 @@ module Fiddle
           value = self.class.new(value)
         end
         if value.is_a?(Fiddle::Pointer)
+          if @resolved_pointer
+            if value.resolved_pointer
+              @resolved_pointer.ffi_ptr = value.resolved_pointer.ffi_ptr
+            else
+              @resolved_pointer.ffi_ptr = value.ffi_ptr
+            end
+          end
           value = value.to_str(args[1])
         end
 
         @ffi_ptr.put_bytes(args[0], value, 0, args[1])
       elsif args.size == 1
         if value.is_a?(Fiddle::Pointer)
+          if @resolved_pointer
+            if value.resolved_pointer
+              @resolved_pointer.ffi_ptr = value.resolved_pointer.ffi_ptr
+            else
+              @resolved_pointer.ffi_ptr = value.ffi_ptr
+            end
+          end
           value = value.to_str(args[0] + 1)
         else
           value = value.chr
@@ -332,6 +348,7 @@ module Fiddle
       end
       @free = free
       @ffi_ptr = ptr
+      @resolved_pointer = nil
       @freed = false
     end
 
@@ -497,6 +514,7 @@ module Fiddle
     def ref
       cptr = Pointer.malloc(FFI::Type::POINTER.size, RUBY_FREE)
       cptr.ffi_ptr.put_pointer(0, ffi_ptr)
+      cptr.resolved_pointer = self
       cptr.size = 0
       cptr
     end
diff --git a/test/fiddle/test_pointer.rb b/test/fiddle/test_pointer.rb
index 73f163d..ae9545b 100644
--- a/test/fiddle/test_pointer.rb
+++ b/test/fiddle/test_pointer.rb
@@ -193,6 +193,19 @@ module Fiddle
       assert_nil(ptr <=> 10, '10 should not be comparable')
     end
 
+    def test_ref
+      ptr = Fiddle::Pointer["hello"]
+      ref = ptr.ref
+      assert_equal(0, ref.size)
+      assert_equal(Fiddle::RUBY_FREE, ref.free)
+      assert_equal(ptr, ref.ptr)
+
+      ptr2 = Fiddle::Pointer["world"]
+      ptr.ref[0, Fiddle::SIZEOF_VOIDP] = ptr2.ref
+      assert_equal("world", ptr.to_s)
+      assert_equal(ptr.to_i, ptr2.to_i)
+    end
+
     def test_ref_ptr
       if ffi_backend?
         omit("Fiddle.dlwrap([]) isn't supported with FFI backend")
```

kou:
Thanks.

ankane:
Big thanks for all of the reviews. That's all of the inconsistencies I'm aware of right now with the FFI backend (and no longer seeing crashes), but will submit more PRs if others come up.

kou:
OK. I'll release a new version.

ankane:
Awesome!

ankane:
Here's the list of gem names:

- mitie
- blingfire
- cmfrec
- gslr
- tsne
- scs
- thundersvm
- highs
- osqp
- glpk
- argon2-kdf
- nlopt
- rcf
- rdkit
- clp
- cbc

eregon:
Ooh, that's nice, thank you for all this testing!

If truffleruby is added in the CI of those gems I'm happy to add these gems to https://github.com/eregon/truffleruby-gem-tracker which lets TruffleRuby devs know quickly if something would regress / checks all these CI jobs are passing.

ankane:
I'm not sure if I'll add it CI (just testing locally for now), but that's good to know.

fwiw, around 70% of my gems currently have a passing test suite with TruffleRuby. The others are failing for one of the following reasons:

1. numo-narray - https://github.com/oracle/truffleruby/issues/3751#issuecomment-2819693184
2. Rice - https://github.com/oracle/truffleruby/issues/3357#issuecomment-2819738306
3. Magnus - https://github.com/oracle/truffleruby/issues/3396#issuecomment-2580778963
4. FFI struct by value support - https://github.com/oracle/truffleruby/issues/3835

PR #189, opened 2025-05-05
Fix "ambiguous first argument" warning in test suite

casperisfine:
```
test/fiddle/test_pointer.rb:288: warning: ambiguous first argument;
put parentheses or a space even after '-' operator
```

I noticed this in one of my ruby builds: https://github.com/ruby/ruby/actions/runs/14838247932/job/41654544867?pr=13159#step:8:1818

cc @kou @ankane

### DIFF
--- a/test/fiddle/test_pointer.rb
+++ b/test/fiddle/test_pointer.rb
@@ -281,7 +281,7 @@ module Fiddle
 
     def test_size
       Pointer.malloc(4, Fiddle::RUBY_FREE) do |ptr|
-        assert_equal 4, ptr.size
+        assert_equal(4, ptr.size)
       end
       assert_equal(0, Pointer.new(0).size)
       assert_equal(0, Pointer.new(0).ref.size)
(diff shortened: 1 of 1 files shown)


test/fiddle/test_pointer.rb
```diff
@@ -285,7 +285,7 @@ def test_size
       end
       assert_equal 0, Pointer.new(0).size
       assert_equal 0, Pointer.new(0).ref.size
```
kou, line 287: Let's not omitting parenthesis for all assertions:

```suggestion
      assert_equal(0, Pointer.new(0).size)
      assert_equal(0, Pointer.new(0).ref.size)
```

test/fiddle/test_pointer.rb
```diff
@@ -283,9 +283,9 @@ def test_size
       Pointer.malloc(4, Fiddle::RUBY_FREE) do |ptr|
         assert_equal 4, ptr.size
```
kou, line 284:
```suggestion
        assert_equal(4, ptr.size)
```

kou:
Good catch!

PR #191, opened 2025-09-17
CI: Add GitHub Actions ppc64le/s390x cases

junaruga:
This PR is working in progress to add GitHub Actions Ubuntu 24.04 ppc64le/s390x cases. This is one of the tasks that we add the cases in some ruby/* repositories. The task is tracked at https://github.com/IBM/actionspz/issues/4#issuecomment-3109205041.

Maybe, this PR's change will be for only `.github/workflows/ci.yml`, while you see other changes in other files temporarily to work on this PR easily for now. I needed to sent a PR to this repository to test the ppc64le/s390x CI pipelines, rather than to test on my fork repository because GitHub Actions Ubuntu 24.04 ppc64le/s390x cases don't work in fork repositories unlike GitHub Actions x86_64/arm64 pipelines. 

I add the new job "host-ibm", because there are ppc64le/s390x specific logic, `if: github.repository == 'ruby/fiddle'` and  "Set up Ruby" logic installing Ruby apt package. The `ruby/setup-ruby` doesn't support the ppc64le/s390x cases as far as I know.

As a reference, we have applied the following code to a few ruby/* repositories.

* ruby/ruby: https://github.com/ruby/ruby/blob/master/.github/workflows/ubuntu-ibm.yml
* ruby/prism: https://github.com/ruby/prism/blob/07de61d4702d12c985d0a708445cf6f3e1935989/.github/workflows/main.yml#L81-L100
* ruby/psych: https://github.com/ruby/psych/blob/1d9c52706b2572a37248a1bdf3f83cc6ef07fd83/.github/workflows/test.yml#L64-L86
* ruby/zlib: https://github.com/ruby/zlib/blob/b2ec227080d79fb16e7df2659c8fe943b5a9452d/.github/workflows/test.yml#L45-L64

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,40 @@ jobs:
           cd tmp
           ruby test/run.rb
 
+  host-ibm:
+    if: github.repository == 'ruby/fiddle'
+    name: ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-24.04-ppc64le
+          - ubuntu-24.04-s390x
+
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Set up Ruby
+        run: |
+          sudo apt-get update
+          sudo apt-get install ruby-full bundler libffi-dev
+
+      - run: sudo bundle install --jobs $(nproc)
+
+      - run: rake compile
+
+      - run: ruby -Ilib test/run.rb
+
+      - run: sudo rake install
+
+      - name: Run test against installed gem
+        run: |
+          ruby -run -e mkdir -- -p tmp/
+          ruby -run -e cp -- -pr test/ tmp/
+          cd tmp
+          ruby test/run.rb
+
   docker:
     name: >-
       ${{ matrix.service }}


junaruga:
This PR is ready for review. A maintainer working on this repository, could you review this PR? Thank you.

.github/workflows/ci.yml
```diff
@@ -77,6 +77,41 @@ jobs:
...
+          - ubuntu-24.04-ppc64le
+          - ubuntu-24.04-s390x
+
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Set up Ruby
+        run: |
+          sudo apt-get update
+          sudo apt-get install ruby-full bundler libffi-dev
```
kou, line 98: Can we use `ruby/setup-ruby@v1` on `ubuntu-24.04-{ppc64le,s390x}`?

junaruga: Unfortunately, we cannot use the ruby/setup-ruby for ubuntu-24-04-{ppc64le,s390x} for now, seeing the following logic.

https://github.com/ruby/setup-ruby/blob/3fee6763234110473bd57dd4595c5199fce2c510/common.js#L173-L181

We need to update the ruby/setup-ruby like following case for arm64.

https://github.com/ruby/setup-ruby/commit/fbaef136828f75c08f5f5db5f3f289ee30044a4e

I will work for ruby/setup-ruby to support `ubuntu-24-04-{ppc64le,s390x}`.

junaruga: @kou Do you want to postpone to merge this PR, until the ruby/setup-ruby supports ppc64le/s390x?

eregon: What are `ubuntu-24-04-{ppc64le,s390x}`, are they self-hosted runners?
I don't see them as GitHub-hosted runners at https://docs.github.com/en/actions/reference/runners/github-hosted-runners at least.
In general, ruby/setup-ruby can only support GitHub-hosted runners.

eregon: According to https://github.com/IBM/actionspz/blob/main/docs/onboarding.md they are self-hosted.

Also based on https://github.com/IBM/actionspz/issues/4#issuecomment-3076501147 I worry to have issues with limited availability of these runners, and e.g. I wouldn't want them to make the total CI time longer, or have to wait a long time to get such a runner when building some new Ruby version.
Also we already have many platforms and it's causing problems like https://github.com/ruby/setup-ruby/issues/752#issuecomment-3299620427

Nevertheless, feel free to open an issue or discussion at https://github.com/ruby/setup-ruby about this, it's a better place to discuss that.

.github/workflows/ci.yml
```diff
@@ -77,6 +77,41 @@ jobs:
...
+
+      - run: sudo bundle install --jobs 2
+
+      - run: rake compile
```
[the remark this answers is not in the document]
junaruga: The `rake compile` without `bundle exec` is to align the current style of the other parts. Below is the current status in the latest master branch. I can send another PR for this change later.

```
$ grep -r 'rake compile' .github/workflows/
.github/workflows/ci.yml:      - run: rake compile
.github/workflows/libffi.yml:        run: rake compile -- --with-libffi-dir=${{ matrix.libffi-prefix }}
.github/workflows/libffi.yml:        run: rake compile -- --with-libffi-source-dir=$(pwd)/libffi-${{ matrix.libffi }}
```

.github/workflows/ci.yml
```diff
@@ -77,6 +77,41 @@ jobs:
           cd tmp
           ruby test/run.rb
 
+  host-ibm:
+    if: github.repository == 'ruby/fiddle'
+    name: >-
+      ${{ matrix.os }}
```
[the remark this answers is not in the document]
junaruga: I applied this change.

junaruga:
Thank you for merging this PR.
I see the commit https://github.com/ruby/fiddle/commit/6368ee0f0f96e731241c42297d862acbadc98132 's commit message including this page's first comment. That was my intention. The first comment was before your reviews. So, the content is outdated for example, "Maybe, this PR's change will be for only `.github/workflows/ci.yml`" was my statement before your review.

junaruga:
> OK. Let's use another PR.

I sent the PR https://github.com/ruby/fiddle/pull/191 for this now.

PR #202, opened 2026-06-06
Don't run TruffleRuby on macOS with Intel CPU on CI

andrykonchin:
Support of macOS + Intel was dropped in the recent TruffleRuby release. So remove a corresponding CI job.

cc @eregon


eregon (approved):
Thanks

eregon:
Fixes this CI failure: https://github.com/ruby/fiddle/actions/runs/25350972762/job/74330240082
cc @kou

kou:
Thanks.

PR #208, opened 2026-07-23
Add `Fiddle::MemoryView.available?`

himura467:
Fiddle had no binding for `rb_memory_view_available_p()`, which callers can use to check whether an object supports the memory view protocol.

- [x] rake test

### DIFF
--- a/test/fiddle/test_memory_view.rb
+++ b/test/fiddle/test_memory_view.rb
@@ -30,6 +30,12 @@ module Fiddle
       end
     end
 
+    def test_available?
+      assert {MemoryView.available?(Pointer["hello world"])}
+      assert {!MemoryView.available?(Fiddle::NULL)}
+      assert {!MemoryView.available?(Object.new)}
+    end
+
     def test_memory_view_from_pointer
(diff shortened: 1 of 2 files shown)


test/fiddle/test_memory_view.rb
```diff
@@ -30,6 +30,12 @@ def test_memory_view_from_unsupported_obj
       end
     end
 
+    def test_available?
+      assert_equal(true, MemoryView.available?(Pointer["hello world"]))
+      assert_equal(false, MemoryView.available?(Fiddle::NULL))
+      assert_equal(false, MemoryView.available?(Object.new))
```
kou: Could you use power assert style for better message on failure?

```suggestion
      assert {MemoryView.available?(Pointer["hello world"])}
      assert {!MemoryView.available?(Fiddle::NULL)}
      assert {!MemoryView.available?(Object.new)}
```

kou (approved):
+1

test/fiddle/test_memory_view.rb
himura467: Fixed.

kou:
Thanks.

PR #212, opened 2026-08-07
Keep the Closure in place for libffi across GC compaction

### LINKED ISSUE #211: Fiddle::Closure segfaults after GC compaction: raw VALUE handed to libffi as closure user-data
## Summary

`Fiddle::Closure` hands libffi the raw `VALUE` of the Closure object as the trampoline's user-data, and casts it back on every invocation. `closure_data_type` declares `.dmark = 0` and no `.dcompact`, so nothing keeps that object in place. If GC compaction relocates the Closure between `initialize` and a later call through the trampoline, libffi passes back a dead address and the process **segfaults**.

## Reproduction

```ruby
require "fiddle"

class Adder < Fiddle::Closure
  def call(a, b) = a + b
end

# Parked off-stack on purpose: a local variable is conservatively pinned by the
# machine-stack scan, which masks the bug.
$holder = [Adder.new(Fiddle::TYPE_INT, [Fiddle::TYPE_INT, Fiddle::TYPE_INT])]

GC.verify_compaction_references(expand_heap: true, toward: :empty)

closure = $holder[0]
func = Fiddle::Function.new(closure.to_i, [Fiddle::TYPE_INT, Fiddle::TYPE_INT], Fiddle::TYPE_INT)
p func.call(40, 2)
```

Actual — 3/3 runs:

```
[BUG] Segmentation fault at 0x0000000000000004
...(with_gvl_callback+0x4c) closure.c:81
c:0003 p:---- s:0014 e:000013 l:y b:0001 CFUNC  :call
```

Expected, and what you get with the `GC.verify_compaction_references` line removed (I ran that as a control): `42`.

`Fiddle::Closure::BlockCaller` and `Fiddle::Importer#bind` fail identically — e.g. passing a `BlockCaller` to libc `qsort` after compaction segfaults at the same address.

## Cause

`ext/fiddle/closure.c` (line numbers from master):

```c
// :60   nothing marks, pins, or relocates the stored copy
.dmark = 0,

// :320  the Closure's own VALUE becomes libffi's user-data
result = ffi_prep_closure_loc(pcl, cif, callback,
                              (void *)(data->self), cl->code);

// :79   cast straight back on every invocation
VALUE self      = (VALUE)x->ctx;
VALUE rbargs    = rb_iv_get(self, "@args");
```

`TypedData_Make_Struct` leaves the Closure movable. Being reachable keeps it alive but not in place; after relocation `x->ctx` refers to whatever now occupies the address, and `rb_iv_get` on it is a wild read.

Worth noting that fiddle already knows about this hazard class: `Fiddle::Pinned` exists precisely because `Fiddle.dlwrap` / `Fiddle::Pointer#to_value` store an unpinned `VALUE` as an integer, and wrapping the object in `Fiddle::Pinned` is the documented mitigation. `Closure` uses the same unsafe primitive *internally*, where a user has no opportunity to apply that mitigation.

Ordinary `GC.compact` does not reliably reproduce it — the object has to actually be relocated, which `verify_compaction_references` forces and which `GC.auto_compact = true` makes possible in the wild.

## Proposed fix

Hand libffi the malloc'd `fiddle_closure *`, which never moves, keep the Closure's `VALUE` in that struct, and mark/relocate it.

```diff
 typedef struct {
+    VALUE self;
     void * code;
     ffi_closure *pcl;
     ffi_cif cif;
     int argc;
     ffi_type **argv;
 } fiddle_closure;

+static void
+closure_mark(void *ptr)
+{
+    fiddle_closure *cl = ptr;
+    rb_gc_mark_movable(cl->self);
+}
+
+static void
+closure_compact(void *ptr)
+{
+    fiddle_closure *cl = ptr;
+    cl->self = rb_gc_location(cl->self);
+}
+
 const rb_data_type_t closure_data_type = {
     .wrap_struct_name = "fiddle/closure",
     .function = {
-        .dmark = 0,
+        .dmark = closure_mark,
         .dfree = dealloc,
-        .dsize = closure_memsize
+        .dsize = closure_memsize,
+        .dcompact = closure_compact
     },

     TypedData_Get_Struct(data->self, fiddle_closure, &closure_data_type, cl);
+
+    /* libffi is handed &cl, which never moves; cl->self is what the
+     * trampoline reads back, so it must be marked and relocated. */
+    RB_OBJ_WRITE(data->self, &cl->self, data->self);

-    VALUE self      = (VALUE)x->ctx;
+    VALUE self      = ((fiddle_closure *)x->ctx)->self;

 #if USE_FFI_CLOSURE_ALLOC
     result = ffi_prep_closure_loc(pcl, cif, callback,
-                                  (void *)(data->self), cl->code);
+                                  (void *)cl, cl->code);
 #else
-    result = ffi_prep_closure(pcl, cif, callback, (void *)(data->self));
+    result = ffi_prep_closure(pcl, cif, callback, (void *)cl);
 #endif
```

I built this against fiddle 1.1.8 and verified it:

- the reproduction above passes 3/3 instead of segfaulting, and the control still passes
- `Closure::BlockCaller` and `Importer#bind` through `qsort` both pass under compaction
- **the upstream test suite is unchanged**: 227 tests, 608 assertions, 0 failures, 1 error, 3 omissions — byte-identical to the same suite run against stock 1.1.8 on this machine (that 1 error is pre-existing, not introduced)

`rb_gc_mark_movable` + `dcompact` avoids pinning; plain `rb_gc_mark` would also close the bug with a smaller diff, at the cost of pinning one object per live Closure.

Happy to send this as a PR with a regression test, if you'd like it in that form.

## Environment

```
fiddle 1.1.8
ruby 4.0.6 (2026-07-14 revision 03b6d3f889) +PRISM [arm64-darwin23]
```

Present on master as of today: `.dmark = 0` at `closure.c:60`, the cast-back at `:79`, and the `ffi_prep_closure_loc` call at `:320-321`.

## Context

Found while auditing this bug class across C extensions after fixing the equivalent problem in sqlite3-ruby (sparklemotion/sqlite3-ruby#723). Same root cause has now been confirmed in psych (ruby/psych#811) and nokogiri (sparklemotion/nokogiri#3665). `ffi` is not affected — it already passes a malloc'd `Closure *` as user-data and relocates the Ruby proc in its compact callbacks, which is the same shape as the fix proposed here.

jeremy:
Closes #211.

## Problem

`Fiddle::Closure` gives libffi the `VALUE` of the Closure object as user data. The trampoline casts that value back on each call.

`closure_data_type` sets `.dmark = 0` and declares no `.dcompact`. Nothing keeps the object in place. GC compaction can move the Closure. After a move, libffi gives back a dead address, and the process stops with SIGSEGV.

`Fiddle::Closure::BlockCaller` and `Fiddle::Importer#bind` fail in the same way.

## Change

libffi now receives the `fiddle_closure` struct. That memory comes from `xmalloc` and does not move.

The struct holds the Closure `VALUE` in a new `self` member:

- `closure_mark` marks the member with `rb_gc_mark_movable`.
- `closure_compact` updates the member with `rb_gc_location`.
- The trampoline reads `((fiddle_closure *)x->ctx)->self`.

The Closure stays movable. The `ffi` gem uses the same method.

## Test

The new test is `test_call_after_compaction` in `test/fiddle/test_closure.rb`.

The test keeps the Closure in an array. A local variable is pinned by the conservative machine-stack scan. A pinned Closure does not move, and then the test cannot fail. A comment in the test records this reason.

| build | result |
|---|---|
| before this change | SIGSEGV at `0x4` |
| after this change | pass |

The full suite after this change: 242 tests, 670 assertions, 0 failures, 0 errors, 3 omissions.

Environment: ruby 4.0.6 (2026-07-14 revision 03b6d3f889) +PRISM [arm64-darwin23].

## Alternative

`rb_gc_mark` also corrects the fault, and the diff is smaller. But it pins one object for each live Closure. This pull request uses `rb_gc_mark_movable` to prevent the pin. Please tell me if you prefer the smaller diff.


test/fiddle/test_closure.rb
```diff
@@ -172,6 +172,30 @@ def call
       end
     end
 
+    def test_call_after_compaction
+      omit("Need GC.verify_compaction_references") unless
+        GC.respond_to?(:verify_compaction_references)
```
kou: Could you avoid using "modifier unless" if we need multiple lines?

```suggestion
      unless GC.respond_to?(:verify_compaction_references)
        omit("Need GC.verify_compaction_references") 
      end
```

ext/fiddle/closure.c
```diff
@@ -54,12 +55,27 @@ closure_memsize(const void * ptr)
     return size;
 }
 
+static void
+closure_mark(void *ptr)
+{
+    fiddle_closure *closure = ptr;
+    rb_gc_mark_movable(closure->self);
+}
```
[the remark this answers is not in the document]
jeremy: Fixed in f695b26 — both functions now check the pointer.

Please note that I could not make it fail. The garbage collector does not call a mark function with a NULL data pointer. I tested this on the code before the fix: 200 freed closures, still referenced from a global array, then `GC.stress = true`, 20 collections, and `GC.verify_compaction_references`. The process survived every time.

The other mark functions in this extension (`pinned_mark`, `fiddle_memview_mark`) also do not check the pointer. But `closure.c` is the only file that sets the data pointer to NULL, so the check belongs here. A comment records both facts.

test/fiddle/test_closure.rb
jeremy: Fixed in f695b26. I applied your suggestion. Thank you.

kou:
Could you fix CI failures?

ext/fiddle/closure.c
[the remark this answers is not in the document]
kou: You're right. We don't need `NULL` check here.

ext/fiddle/closure.c
```diff
@@ -54,12 +55,45 @@ closure_memsize(const void * ptr)
...
+
+/* Fiddle::Closure#free and initialize_rescue set the data pointer to NULL.
+ * The garbage collector does not call these functions with a NULL pointer,
+ * but this file is the only one in the extension that clears the pointer,
+ * so both functions check it. */
+static void
+closure_mark(void *ptr)
+{
+    fiddle_closure *closure = ptr;
+    if (closure) {
```
kou: Could you remove redundant `NULL` check?

jeremy: Fixed in e06557d — removed both `NULL` checks and the comment that justified them.

kou:
Thanks.

PR #213, opened 2026-08-08
Test that a coerced TYPE_CONST_STRING argument stays in place

jeremy:
## Problem

`TYPE_CONST_STRING` stores a pointer into a Ruby String:

```c
case TYPE_CONST_STRING:
    ...
    else { dst->pointer = rb_string_value_cstr(src); }
```

A String argument is safe. `argv` holds it and keeps it in place.

A `to_str` object is not. `rb_string_value_cstr` writes the coerced String back into `src`. `argv` still holds the original object, not the String. The next loop turn overwrites `src`. After that, only `converted_args` holds the String.

The pointer must stay valid until `ffi_call` returns. `ffi_call` runs under `rb_thread_call_without_gvl`, so another thread can compact the heap during the call. `converted_args` must keep the String **in place**, not only alive.

Master does this. `converted_args` sits in the `ALLOCV` buffer. The GC scans that buffer conservatively, and a conservative scan pins the object.

## Gap

No test covers this.

Before #205, `converted_args` was a Ruby Array. The GC may move an Array's elements, so the String moved while the pointer stayed, and the C function read the vacated slot. #205 replaced the Array with the buffer and fixed the fault as a side effect. Nothing records that the buffer is required.

## Change

1. `test_call_const_string_from_to_str_after_compaction` in `test/fiddle/test_function.rb`.
2. A comment at `converted_args` in `ext/fiddle/function.c` stating the requirement.

The test calls `strcspn` as `[TYPE_CONST_STRING, TYPE_VOIDP] -> TYPE_SIZE_T`. Arg 0 is a `to_str` object returning a fresh 100-byte String. Arg 1 is a `to_ptr` object, which makes Fiddle call `Pointer.[]`; that allocation lets the GC run after the pointer is stored.

The subject holds no `"Z"`; the reject set is `"Z"`. A correct call returns 100. A stale pointer returns 0, because a vacated slot holds zero bytes.

Three details are load-bearing. Remove any one and the test passes on a faulty build; each has a comment:

- Churn is the subject's byte size, half of it dropped. The compactor only evacuates non-full pages.
- `to_ptr` returns a **new** `Pointer` each time. An existing `Pointer` gives Fiddle nothing to allocate.
- The intact-case call runs **after** the loop. A call before it prepares the CIF, and that preparation is part of what opens the window.

## Result

| tree | arm64-darwin, ruby 3.4.10 | aarch64-linux, ruby 3.4.7 |
|---|---|---|
| `d389bbf` (before #205) | fail | fail |
| master | pass | pass |

Three runs per tree on darwin, same result each time. Full suite on master with this change: 242 tests, 668 assertions, 0 failures, 1 error, 3 omissions. The error is `test_nogvl_poll` needing `envutil`; it fails the same way without this change.

Environment: ruby 3.4.10 [arm64-darwin23]; ruby 3.4.7 [aarch64-linux], glibc 2.36, libffi 3.4.4.

## Note

Released versions still have the fault. I measured 1.1.6 and 1.1.8 on both platforms: a two-argument call of this shape returns the wrong result on every iteration under `GC.stress` with `GC.auto_compact`. Master is correct. You may want a release.

I report this as latent. The trigger is the declared argument type plus a `to_str` object in place of a String. In one application that loads fiddle at boot, all 227 bundled gems put every `Fiddle::Function.new` behind a Windows-only branch.

### DIFF
--- a/test/fiddle/test_function.rb
+++ b/test/fiddle/test_function.rb
@@ -179,6 +179,81 @@ module Fiddle
       assert_equal("123", str.to_s)
     end
 
+    # A TYPE_CONST_STRING argument that comes from a to_str object is coerced inside
+    # Fiddle. The caller's argv then holds that object, not the String that the
+    # char * points into, so argv does not keep the String in place. Fiddle keeps the
+    # coerced String in converted_args instead, and the conservative scan of the
+    # ALLOCV buffer pins it there.
+    #
+    # That pin is what this test protects. A converted_args that the garbage
+    # collector can move, such as a Ruby Array, lets compaction move the String after
+    # the char * is stored. The C function then reads the slot the String left.
+    #
+    # The second argument must be neither a String nor a Pointer. Only then does
+    # Fiddle call Pointer.[] on it, and that allocation is what lets the garbage
+    # collector run while the char * is already in place.
+    def test_call_const_string_from_to_str_after_compaction
+      omit("Need CRuby") unless RUBY_ENGINE == "ruby"
+      omit("Need GC.auto_compact=") unless GC.respond_to?(:auto_compact=)
+
+      length = 100
+      subject_class = Class.new do
+        define_method(:to_str) { "Q" * length }
+      end
+      # 4096 bytes keep these bytes in a malloc'd buffer, which compaction does not
+      # move. A short String would hold its bytes in the object itself, and then this
+      # pointer could dangle and report a failure that this test is not about.
+      reject = "Z" + ("\0" * 4095)
+      # to_ptr must build a new Pointer on every call. A Pointer that already exists
+      # gives Fiddle nothing to allocate, the garbage collector then does not run
+      # while the char * is in place, and this test cannot fail.
+      opener_class = Class.new do
+        define_method(:to_ptr) { Pointer[reject] }
+      end
+
+      strcspn = Function.new(@libc["strcspn"],
+                             [TYPE_CONST_STRING, TYPE_VOIDP],
+                             TYPE_SIZE_T)
+      # The subject holds no "Z", so strcspn stops at the terminator and reports the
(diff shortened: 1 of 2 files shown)


test/fiddle/test_function.rb
```diff
@@ -179,6 +179,86 @@ def test_strcpy
...
+          GC.stress = true
+          begin
+            results << strcspn.call(subject_class.new, opener_class.new)
+          ensure
+            GC.stress = false
+            GC.auto_compact = auto_compact
+          end
+        end
+      ensure
+        GC.stress = false
```
kou: Do we need this?

kou (commented):
Is the added test enough simple?

test/fiddle/test_function.rb
jeremy: Fixed — it was there because the toggles sat outside the inner `begin`, so an interrupt between `GC.stress = true` and the `begin` could leak stress mode. Moved the toggles inside the inner `begin` instead, so its `ensure` covers every path, and dropped the outer `begin`/`ensure` entirely.

jeremy:
Simplified the nested `begin`/`ensure` down to one (the outer one only existed to cover the toggles sitting outside the inner `begin`; moving them inside removes the need).

The rest is hard to shrink further because the compaction has to fire *in the middle* of the call — after the `char *` is stored into the args buffer but before `ffi_call` — so `GC.verify_compaction_references` can't be used here. That leaves `GC.stress` + `GC.auto_compact` to make the GC fire on the `Pointer[]` allocation inside the window, and the churn/burst setup to make the compactor actually evacuate the subject's page. Each step carries a comment saying what stops the test from being able to fail without it.

kou:
Thanks.

